PR #77, opened 2018-08-20
Update import paths for checks

ryanall:


cesar-rodriguez:
Thanks for contributing! I'm closing this in favor of #84 as that one passes CI.

PR #269, opened 2020-08-12
wip: update terrascan usage information

williepaul:


cesar-rodriguez:
The -version flag doesn't seem to be working.

PR #270, opened 2020-08-13
added line number and file name output support

williepaul:
policy changes—removed policies with errors (will need to fix).
these errors include things like missing reference IDs and invalid severity strings
record time duration of the scan
fixed Engine.GetResults and updated the interface
reduced the scope of the members of the Engine type
misc code cleanup


williepaul:
The branch name is misleading, but this includes the line/source file changes amongst other things.

PR #308, opened 2020-09-01
Manage relative module path

guilhem:
Our modules are called with:
```
source = "../../modules/mymodule"
```
and fail


kanchwala-yusuf (approved):

kanchwala-yusuf:
@guilhem Thank you for your contribution!

PR #309, opened 2020-09-02
add support to print resource config as an output

kanchwala-yusuf:


cesar-rodriguez:
Just tried this and it works as expected. Would you mind updating the docs explaining the new command?

kanchwala-yusuf:
@cesar-rodriguez , I have update the documentation as well. Can you please verify?

williepaul (approved):

PR #333, opened 2020-09-17
Add check for visibility for github_repository

### LINKED ISSUE #326: Private GitHub repositories are not recognized with version 3.0.0+ of GitHub provider
The latest version of the GitHub provider deprecates the usage of `private` in favor of the `visibility` configuration. Following this advice Terrascan complains with `accurics.gcp.IAM.145` as only the private configuration seems to be recognized. [privateRepoEnabled.rego](https://github.com/accurics/terrascan/blob/master/pkg/policies/opa/rego/gcp/github_repository/privateRepoEnabled.rego) should be extended to also allow the usage of the `visibility = "private"` configuration instead of `private`.

HorizonNet:
Add a check for the visiblity property of the github_repository
resource from the GitHub provider. It checks if the visibility was not
set to "private". This complies with the latest version of the provider
(3.0.0+), which switches to visibility in favor of the private property.
Left the private property in for backward compatibility.

Resolves #326

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,9 @@ require (
 	go.uber.org/zap v1.10.0
 	golang.org/x/net v0.0.0-20200822124328-c89045814202 // indirect
 	golang.org/x/sys v0.0.0-20200814200057-3d37ad5750ed // indirect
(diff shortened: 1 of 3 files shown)


williepaul (approved):
Nice job!

PR #347, opened 2020-10-01
Add support for colorized output

acc-jon:
Add a command-line option to specify whether output should be colored (default="auto").  In auto mode, colors will be enabled if the output will be written to a terminal.  The coloring is applied after the output has been marshalled, to avoid corrupting any output data.  Add --use-colors=false to force colors off.

Note that coloring is only applied to JSON and YAML output, because coloring XML is not helpful.  XML is already too dense to read easily.

Default colors are built in, but are user configurable via a JSON file.  Example in pkg/termcolor/color_patterns.json.sample.  Enable the custom colors with an environment variable, TERRASCAN_COLORS_FILE=<absolute-path-to-color_patterns.json>


cesar-rodriguez (approved):
Thanks!

PR #350, opened 2020-10-10
Add new policy for checking insecure_ssl on github_organization_webhook

### LINKED ISSUE #339: Add policy for checking insecure_ssl configuration for github_organization_webhook in GitHub provider
The GitHub provider has a resource [github_organization_webhook](https://www.terraform.io/docs/providers/github/r/organization_webhook.html), which can be used to setup a webhook for an organization. In the configuration it allows several options. One of the them is `insecure_ssl`, which is [not recommended](https://developer.github.com/v3/orgs/hooks/#create-an-organization-webhook) to be set to `true`. The new policy for the `github` policy set should check this configuration and report a violation if this option was set to `true`.

HorizonNet:
Add a new policy to the GitHub provider to check if insecure_ssl was
set to true on the github_organization_webhook resource.

Resolves #339

### DIFF
--- a/docs/policies/github.md
+++ b/docs/policies/github.md
@@ -5,3 +5,7 @@
 | Identity & Access Management | github | MEDIUM | Repository is Not Private. | accurics.github.IAM.1 |
 
 
+### github_organization_webhook
+| Category | Resource | Severity | Description | Reference ID |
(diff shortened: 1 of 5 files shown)


cesar-rodriguez (approved):

cesar-rodriguez:
Thanks!

PR #354, opened 2020-10-16
Allow configuration of global policy config, fix some typos

acc-jon:


kanchwala-yusuf (changes requested):
Hi @acc-jon,

We already have a config file in terrascan. You can check the sample config file at https://github.com/accurics/terrascan/blob/master/config/terrascan.toml.

We have adopted the `toml` format for config file, as it is intuitive, easy to use and read. 

I see that you expect the Global config file indicated by `TERRASCAN_CONFIG` to be a json file. This creates two types of config files within terrascan. 

Hence, I would request you revise this once again.

pkg/config/configfile.go
```diff
@@ -0,0 +1,60 @@
...
+	// empty config file path
+	if configFile == "" {
+		zap.S().Debug("no config file specified")
+		return nil, nil
+	}
+
+	// check if file exists
+	_, err := os.Stat(configFile)
+	if err != nil {
+		zap.S().Errorf("Can't find '%s'", configFile)
```
kanchwala-yusuf, line 47: Small suggestion to not have error strings capitalized: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

kanchwala-yusuf (approved):

PR #372, opened 2020-11-06
Address #365 by properly handling submodule path

### LINKED ISSUE #365: Deep modules location mis-proccessed.
* terrascan version: Latest (2c964d54cb798d1dd5de05836af13ca5684ffb7c)
* Operating System: OS X (Darwin MacBook-Pro.local 18.7.0 Darwin Kernel Version 18.7.0: Mon Aug 31 20:53:32 PDT 2020; root:xnu-4903.278.44~1/RELEASE_X86_64 x86_64)

### Description

I have  a situation where I am using several modules together.  For simplicity, we will just call them m1, m2, and m3. m1 is consumed from my main terraform template and it consumes m2 and m3. When I try and run terrascan, it incorrectly parsers the directory in the module as relative to absRootDir instead of relative to the location of m1.

The file system would look like this
/tf/template.tf
/tf/modules/m1/main.tf
/tf/modules/m2/main.tf

/tf/template.tf  contains a call like:
```
module "m1" {
    source "./modules/m1/"
}
```

/tf/modules/m1/main.tf contains a call like
```
module "m2" {
    source "../m2/"
}

That works just fine in terraform and it parses them correctly. 

```
### What I Did
When I run terrascan on the directory though, I get the following error
```
$ terrascan scan -t aws 
2020-10-29T10:08:46.046-0700	error	v12/load-dir.go:110	failed to build unified config. errors:
<nil>: Failed to read module directory; Module directory /tf/m2 does not exist or cannot be read.
```

I tossed some debugging in to see how it is parsing it internally into the module walker function:

```
2020/10/29 10:16:11 req.Path.String(), is m1
2020/10/29 10:16:11 req.SourceAddr, is ./modules/m1
2020/10/29 10:16:11 Created path to module as  /tf/modules/m1
2020/10/29 10:16:11 processing local module "./modules/m1"
2020/10/29 10:16:11 req.Path.String(), is m1.m2
2020/10/29 10:16:11 req.SourceAddr, is ../m2
2020/10/29 10:16:11 Created path to module as  /tf/m3
2020/10/29 10:16:11 processing local module "../m2"
```
The code added at the end of the if statement at line 84 in pkg/iac-providers/terraform/v12/load-dir.go to generate the above is:
```
log.Println("req.Path.String(), is", req.Path.String())
log.Println("req.SourceAddr, is", req.SourceAddr)
log.Println("Created path to module as ", pathToModule)
log.Printf("processing local module %q", req.SourceAddr)
```

I think the mistake is this code here (still in the if statement that starts at 84):
```
pathArr := strings.Split(req.Path.String(), ".")
pathArr = pathArr[:len(pathArr)-1]
pathToModule = filepath.Join(absRootDir, filepath.Join(pathArr...), req.SourceAddr)
```

Instead of looking at the req.Path.String, it should be looking at req.Parent.SourceAddr.  If that were used instead, it would result in pathToModules containing the following values:
```
absRootDir = "/tf/"
req.Parrent.SourceAddr = "./modules/m1"
req.SourceAddr = "../m2"
```

which would get combined with the filepath.Join to result in 
```
pathToModule = "/tf/modules/m2"
```

acc-jon:
This fixes #365 by properly handling sub-module path.

### DIFF
--- a/pkg/iac-providers/terraform/v12/load-dir.go
+++ b/pkg/iac-providers/terraform/v12/load-dir.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"github.com/accurics/terrascan/pkg/iac-providers/output"
 	"github.com/accurics/terrascan/pkg/utils"
@@ -84,9 +83,7 @@ func (*TfV12) LoadIacDir(absRootDir string) (allResourcesConfig output.AllResour
 			if isLocalSourceAddr(req.SourceAddr) {
 				// determine the absolute path from root module to the sub module
 				// using *configs.ModuleRequest.Path field
-				pathArr := strings.Split(req.Path.String(), ".")
-				pathArr = pathArr[:len(pathArr)-1]
-				pathToModule = filepath.Join(absRootDir, filepath.Join(pathArr...), req.SourceAddr)
+				pathToModule = filepath.Join(absRootDir, req.Parent.SourceAddr, req.SourceAddr)
 				zap.S().Debugf("processing local module %q", req.SourceAddr)
 			} else {
 				// temp dir to download the remote repo
--- a/pkg/iac-providers/terraform/v12/load-dir_test.go
+++ b/pkg/iac-providers/terraform/v12/load-dir_test.go
@@ -91,6 +91,13 @@ func TestLoadIacDir(t *testing.T) {
 			tfv12:       TfV12{},
 			wantErr:     nil,
 		},
+		{
+			name:        "nested module directory",
+			tfConfigDir: "./testdata/deep-modules",
+			tfJSONFile:  "./testdata/tfjson/deep-modules.json",
+			tfv12:       TfV12{},
+			wantErr:     nil,
+		},
 	}
 
 	for _, tt := range table2 {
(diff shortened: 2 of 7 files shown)


kanchwala-yusuf (approved):

pkg/iac-providers/terraform/v12/load-dir.go
```diff
@@ -84,9 +83,7 @@ func (*TfV12) LoadIacDir(absRootDir string) (allResourcesConfig output.AllResour
 			if isLocalSourceAddr(req.SourceAddr) {
 				// determine the absolute path from root module to the sub module
 				// using *configs.ModuleRequest.Path field
-				pathArr := strings.Split(req.Path.String(), ".")
-				pathArr = pathArr[:len(pathArr)-1]
-				pathToModule = filepath.Join(absRootDir, filepath.Join(pathArr...), req.SourceAddr)
+				pathToModule = filepath.Join(absRootDir, req.Parent.SourceAddr, req.SourceAddr)
```
salexpdx, line 86: I'm concerned that this only moves the issue one level deeper.  If you go back to https://github.com/salexpdx/terrascan/commit/89b06f6b18881274ff7175bf418cf7a463077fa0#diff-ed0a902412e1c2c9cca7b9c8d22f14edcd3711f1535d0bc128e801e3ac300c84R95, you can see that i looped through.  That is done so that if there is more than one layer of sub-modules, it will parse through all the way.

PR #377, opened 2020-11-11
Adds support for Helm v3

### LINKED ISSUE #353: Add support for Helm
### Description

We should add support for [Helm](https://helm.sh/) as an IaC provider.

williepaul:
- supports loading of helm v3 charts via the IaC dir command (IaC file is invalid)
- adds default IaC versions for all IaC types
- adds unit tests
- adds ability to properly specify default IaC version for a given IaC type

Closes #353


pkg/iac-providers/helm/v3/load-dir.go
```diff
@@ -0,0 +1,289 @@
...
+	errBadChartVersion   = fmt.Errorf("bad chart version in Chart.yaml")
+)
+
+// LoadIacDir loads all helm charts under the specified directory
+func (h *HelmV3) LoadIacDir(absRootDir string) (output.AllResourceConfigs, error) {
+
+	allResourcesConfig := make(map[string][]output.ResourceConfig)
+
+	// find all Chart.yaml files within the specified directory structure
+	fileMap, err := utils.FindFilesBySuffix(absRootDir, []string{helmChartFilename})
```
kanchwala-yusuf, line 50: Can the helm char file also have `.yml` extenstion? In that case, `FindFilesBySuffix` will return an empty fileMap

williepaul: The helm chart file only supports Chart.yaml as far as I can tell. I added .yml at first because rancher uses it, but figured we would probably need to support rancher in a separate iac-type, so I reverted the change.

pkg/iac-providers/helm/v3/load-dir.go
```diff
@@ -0,0 +1,289 @@
...
+		if err != nil {
+			continue
+		}
+		iacDocuments = append(iacDocuments, renderedCharts...)
+	}
+	return iacDocuments, chartMap, err
+}
+
+func (h *HelmV3) getHelmTemplateExtensions() []string {
+	return []string{"yaml", "tpl"}
```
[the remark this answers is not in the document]
williepaul: Agreed--done.

pkg/iac-providers/helm/v3/load-dir.go
williepaul: Confirmed this is indeed the case in the helm source (only looks for "Chart.yaml")

cesar-rodriguez (approved):

kanchwala-yusuf (approved):

PR #385, opened 2020-11-14
fixes dockerfile build issues

williepaul:
- disables CGO to prevent any dependencies from failing the build
- the GOARCH variable was named incorrectly and not being picked up during the docker build
- locked both the alpine and golang versions for the docker build

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,8 +1,9 @@
 # -------- builder stage -------- #
-FROM golang:alpine AS builder
+FROM golang:1.15.5-alpine3.12 AS builder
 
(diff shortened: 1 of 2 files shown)


build/Dockerfile
```diff
@@ -1,8 +1,9 @@
 # -------- builder stage -------- #
-FROM golang:alpine AS builder
+FROM golang:1.15.5-alpine3.12 AS builder
```
kanchwala-yusuf, line 2: Can we keep base image as it is? Updating the golang docker image tag every time can become a maintenance overhead.

cesar-rodriguez (approved):

Issue #330, opened 2020-09-16
Why accurics.gcp.IAM.104 suggests enabling a client certificate?


ferrarimarco:
* terrascan version: v1.1.0
* Operating System: Ubuntu 20.04.1

### Description

[Hardening your cluster security](https://cloud.google.com/kubernetes-engine/docs/how-to/hardening-your-cluster#restrict_authn_methods) recommends to disable client certificates. Why is Terrascan suggesting the opposite?

Thanks!

williepaul:
Hello @ferrarimarco,

This one is a bug; from the rule name "clientCertificateEnabled", it seems the intention was to check if the client cert is enabled (to flag it), but the rule is doing the opposite and triggering if it's false instead. Thanks for the bug report!

lucas-giaco:
Hi folks, I'm facing the same issue. 
There's an ETA for this bug to be fixed? or there's any way we could help to fix it?

Thanks!

cesar-rodriguez:
Resolved with #402

PR #402, opened 2020-11-22
accurics.gcp.IAM.104 Fire rule when client certificate is enabled

lucas-giaco:
Hi folks, this is my first contribution to the project and my first time with rego rules, so thanks in advance for your patience.
I'm trying to fix the clientCertificateEnabled rule to be fired only the client certificate is enabled.

This PR should solve #330


cesar-rodriguez (approved):
Thanks for your contribution!

PR #427, opened 2020-12-09
set console mode on windows so colors render

acc-jon:


pkg/cli/color-console_windows.go
```diff
@@ -0,0 +1,36 @@
...
+
+package cli
+
+import (
+	"golang.org/x/sys/windows"
+	"os"
+)
+
+// In order for colored output to work on Windows, we need to explicitly
+// enable the virtual terminal processing console mode.  This is needed for
```
williepaul, line 27: Hey Jon,

One suggestion--can we detect if virtual terminal processing console mode is enabled, and either switch to no-color automatically, issue a warning, or both?

acc-jon: Hi Willie, the PreRun func for scanCmd in cli/scan.go will disable colored output mode if the terminal is not a tty.

This code just ensures that Windows users will be able to see colored output.  The default mode is disabled, probably for historical reasons that are no longer relevant.

Note that Terrascan uses third-party packages like zap that also expect a color-capable terminal.  If we wanted to disable colors then we would need to figure out how/if those could be forced to a non-color mode.

williepaul (approved):

PR #428, opened 2020-12-10
Add policy AC-K8-NS-SE-M-0188 for CVE-2020-8554

gauravgogia-accurics:


pkg/policies/opa/rego/k8s/kubernetes_service/cve_2020_8554/ensurePrivateIP.rego
```diff
@@ -0,0 +1,16 @@
+package accurics
+
+{{.prefix}}{{.name}}{{.suffix}}[service.id] {
+    service := input.{{.resource_type}}[_]
+    service.config.kind == "Service"
```
williepaul, line 5: Looks like this check is unnecessary--kubernetes_service resource_type implies this kind

pkg/policies/opa/rego/k8s/kubernetes_service/cve_2020_8554/ensurePrivateIP.rego
```diff
@@ -0,0 +1,16 @@
...
+    type_check(service.config.spec)
```
williepaul, line 6: IMO we should keep our checks as simple as possible. it looks like you're just trying to do the following:
service.config.spec.type == "ClusterIP"
not service.config.spec.type
not service.config.spec.externalIPs

Do we also need a check for service.config.spec.clusterIPs?

pkg/policies/opa/rego/k8s/kubernetes_service/cve_2020_8554/ensurePrivateIP.rego
gauravgogia-accurics: I see, will remove that check

pkg/policies/opa/rego/k8s/kubernetes_service/cve_2020_8554/ensurePrivateIP.rego
gauravgogia-accurics: Service type may or may not be set. If its set then I want to if its **ClusterIP**

williepaul (approved):

PR #431, opened 2020-12-12
bash output improvements

patilpankaj212:
- implement default human readable output for terrascan scan
- unit tests for output writers
- change color for "File" in scan output as it is not visible on white background


pkg/results/types.go
```diff
@@ -59,3 +75,34 @@ func (vs ViolationStore) Add(extra ViolationStore) ViolationStore {
 
 	return vs
 }
+
+// NewDefaultScanResult will initialize DefaultScanResult
+func NewDefaultScanResult(iacType, iacFilePath, iacDirPath string, totalPolicyCount int, verbose bool, violationStore ViolationStore) *DefaultScanResult {
```
kanchwala-yusuf, line 80: I think, Scan summary can become part of the `ViolationStore` struct itself. Scan summary is useful in all the formats. Doing so should eliminate the need to have the function `NewDefaulScanResult`.

Does it make sense?

kanchwala-yusuf (changes requested):

pkg/cli/scan.go
```diff
@@ -60,6 +60,9 @@ var (
 	// UseColors indicates whether to use color output
 	UseColors bool
 	useColors string // used for flag processing
+
+	// Verbose indicates whether to display all fields in default human readlbe output
+	Verbose bool
```
kanchwala-yusuf, line 65: usually `Verbose` or `-v` refers to verbosity in log messages. Would it make sense to change this field name to something else?

pkg/results/types.go
patilpankaj212: Yes, we would not need the DefaultScanResult struct if we add that information to ViolationStore struct. But the idea was not to make any change in the existing output.

pkg/runtime/executor.go
```diff
@@ -131,3 +131,13 @@ func (e *Executor) Execute() (results Output, err error) {
 	// successful
 	return results, nil
 }
+
+// GetTotalPolicyCount will return the total count of policies in all engines
+func (e Executor) GetTotalPolicyCount() int {
```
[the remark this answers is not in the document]
williepaul: Absolutely, this should go into the policy package. There is some code that may help with this, which is a refactor of the policy engine code, but it has not been completed yet. I will try to get to this soon. But in the meantime, this code itself should be placed in the policy package itself.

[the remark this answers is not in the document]
patilpankaj212: In the policy package, added a GetRuleCount() method on engine. In the runtime, the executor has the list of policy engines and it will be able to get the total rule count.

pkg/cli/scan.go
patilpankaj212: I had kept it --show-details. Changed it to verbose after discussions with Amir.

pkg/cli/register.go
```diff
@@ -56,7 +56,7 @@ func setDefaultCommandIfNonePresent() {
 func Execute() {
 	rootCmd.PersistentFlags().StringVarP(&LogLevel, "log-level", "l", "info", "log level (debug, info, warn, error, panic, fatal)")
 	rootCmd.PersistentFlags().StringVarP(&LogType, "log-type", "x", "console", "log output type (console, json)")
-	rootCmd.PersistentFlags().StringVarP(&OutputType, "output", "o", "yaml", "output type (json, yaml, xml)")
+	rootCmd.PersistentFlags().StringVarP(&OutputType, "output", "o", "human", "output type (human, json, yaml, xml)")
```
williepaul, line 59: Per this change, human-readable will be the default output. Is this the desired default output format?

pkg/cli/run.go
```diff
@@ -67,9 +74,19 @@ func Run(iacType, iacVersion string, cloudType []string,
 	outputWriter := NewOutputWriter(useColors)
 
 	if configOnly {
+		// human readable output doesn't support --config-only flag
+		// if --config-only flag is set, then display json output
+		if strings.EqualFold(format, humanOutputFormat) {
+			format = jsonOutputFormat
+		}
```
williepaul, line 81: If not supported, we may just opt to return an error and require the user to specify a valid output (-o json, yaml, etc)

pkg/runtime/executor.go
```diff
@@ -131,3 +131,13 @@ func (e *Executor) Execute() (results Output, err error) {
...
+	policyCount := 0
+	for _, engine := range e.policyEngine {
+		opaEngine := engine.(*opa.Engine)
```
williepaul, line 139: The policy engine is not guaranteed to be OPA. Although it is the only engine we support now, it's not a safe assumption to make here... But since this function belongs in the policy package anyway, you should just be able to pull the total directly from the engine.

williepaul (changes requested):

pkg/cli/register.go
patilpankaj212: Yes, this is as per the discussions with Amir.

pkg/cli/run.go
patilpankaj212: The reason the config only output is not supported is because the default scan result doesn't not hold that information. If we move the summary to ViolationStore, then we can support display the resource config with this format too. I will talk with Amir.

pkg/cli/run.go
```diff
@@ -28,10 +29,14 @@ import (
...
 
+const (
+	humanOutputFormat = "human"
+)
+
 // Run executes terrascan in CLI mode
 func Run(iacType, iacVersion string, cloudType []string,
 	iacFilePath, iacDirPath, configFile string, policyPath []string,
-	format, remoteType, remoteURL string, configOnly, useColors bool) {
+	format, remoteType, remoteURL string, configOnly, useColors, verbose bool) {
```
williepaul, line 39: Hey Pankaj,

I misunderstood here--I was assuming that if we're now including the other details within the violation store, that there is no longer need for this verbose flag. Can we remove this?

patilpankaj212: We would still need to have the verbose flag for the default human readable output.

pkg/results/types.go
```diff
@@ -32,30 +38,59 @@ type Violation struct {
...
 }
+
+// AddSummary will update the summary with remaining details
+func (vs *ViolationStore) AddSummary(iacType, iacFilePath, iacDirPath string) {
+	if iacType == "" {
+		// the default scan type is terraform
+		vs.Summary.IacType = "terraform"
+	} else {
+		vs.Summary.IacType = iacType
+	}
```
williepaul, line 82: Is this needed? AddSummary() is called late (after the engine has run), so I was expecting iacType to have some value populated already.

pkg/results/types.go
```diff
@@ -32,30 +38,59 @@ type Violation struct {
...
+	vs.Summary.MediumCount += extra.Summary.MediumCount
+	vs.Summary.HighCount += extra.Summary.HighCount
+	vs.Summary.ViolatedPolicies += extra.Summary.ViolatedPolicies
+	vs.Summary.TotalPolicies += extra.Summary.TotalPolicies
 
 	return vs
 }
+
+// AddSummary will update the summary with remaining details
+func (vs *ViolationStore) AddSummary(iacType, iacFilePath, iacDirPath string) {
```
williepaul, line 76: I think it's preferable to specify the "resource path" directly in the arguments, rather than have iacFile/iacDir. The violation store shouldn't have to know about these details.

Ideally, AddSummary() would just include the bare minimum--
vs.Summary.IacType = iacType
vs.Summary.ResourcePath = resourcePath
vs.Summary.Timestamp = ...

pkg/results/types.go
patilpankaj212: Right !! this is not required. I will remove this check.

williepaul (approved):

pkg/cli/scan.go
williepaul: Per Amir, verbose is ok.

pkg/results/types.go
patilpankaj212: made the change

pkg/results/types.go
patilpankaj212: made the change

pkg/runtime/executor.go
patilpankaj212: This change was removed after adding summary to ViolationStore

williepaul (dismissed):

pkg/cli/run.go
patilpankaj212: done this change. Error will be displayed to user if --config-only flag is used with default human readable output.

pkg/cli/run.go
```diff
@@ -67,9 +74,19 @@ func Run(iacType, iacVersion string, cloudType []string,
...
 		writer.Write(format, results.ResourceConfig, outputWriter)
 	} else {
-		writer.Write(format, results.Violations, outputWriter)
+		if strings.EqualFold(format, humanOutputFormat) {
+			defaultScanResult := result.NewDefaultScanResult(iacType, iacFilePath, iacDirPath, executor.GetTotalPolicyCount(), verbose, *results.Violations.ViolationStore)
+			writer.Write(format, defaultScanResult, outputWriter)
+		} else {
+			writer.Write(format, results.Violations, outputWriter)
+		}
 	}
```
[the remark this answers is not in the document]
williepaul: I need to complete the policy engine refactoring--will address this at that point.

kanchwala-yusuf (approved):

PR #434, opened 2020-12-17
Add support to extract rules to skip from terraform comments

kanchwala-yusuf:
This PR adds support for extracting rule ids to skip policy evaluation. These rules can be set as a comma separated list of rule ids in the comments of terraform resource config in the following pattern:
```hcl
#ts:skip=AWS.S3Bucket.DS.High.1041, AWS.S3Bucket.DS.High.1042
```

Example of a terraform resource config with rules to skip policy evaluation:
```hcl
resource "aws_s3_bucket" "bucket2" {
   #ts:skip=AWS.S3Bucket.DS.High.1041
   region        = var.region
   bucket        = local.bucket_name
   #ts:skip=AWS.S3Bucket.DS.High.1044,AWS.S3Bucket.DS.High.1045
   force_destroy = true
   #ts:skip= AWS.S3Bucket.DS.High.1046   ,   AWS.S3Bucket.DS.High.1047
   acl           = "public-read"
 }
```

### DIFF
--- a/pkg/iac-providers/terraform/v12/resource.go
+++ b/pkg/iac-providers/terraform/v12/resource.go
@@ -19,12 +19,18 @@ package tfv12
 import (
 	"fmt"
 	"io/ioutil"
+	"regexp"
+	"strings"
 
+	"github.com/accurics/terrascan/pkg/iac-providers/output"
 	"github.com/hashicorp/hcl/v2/hclsyntax"
 	hclConfigs "github.com/hashicorp/terraform/configs"
 	"go.uber.org/zap"
+)
 
-	"github.com/accurics/terrascan/pkg/iac-providers/output"
+var (
+	skipRulesPattern = regexp.MustCompile(`#ts:skip=\s*(([A-Za-z0-9]+\.?){5})(\s*,\s*([A-Za-z0-9]+\.?){5})*`)
+	skipRulesPrefix  = "#ts:skip="
 )
 
 // CreateResourceConfig creates output.ResourceConfig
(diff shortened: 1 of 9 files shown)


pkg/iac-providers/terraform/v12/resource.go
```diff
@@ -49,15 +55,40 @@ func CreateResourceConfig(managedResource *hclConfigs.Resource) (resourceConfig
...
 
 	// successful
 	zap.S().Debugf("created resource config for resource '%s', file: '%s'", resourceConfig.Name, resourceConfig.Source)
 	return resourceConfig, nil
 }
+
+// getSkipRules returns a list of rules to be skipped. The rules to be skipped
+// can be set in terraform resource config with the following comma separated pattern:
+// #ts:skip=AWS.S3Bucket.DS.High.1043, AWS.S3Bucket.DS.High.1044
+func getSkipRules(body string) []string {
```
williepaul, line 75: This could be useful for other IaC... Maybe place in utils for now?

pkg/iac-providers/terraform/v12/resource.go
```diff
@@ -19,12 +19,18 @@ package tfv12
...
 
+	"github.com/accurics/terrascan/pkg/iac-providers/output"
 	"github.com/hashicorp/hcl/v2/hclsyntax"
 	hclConfigs "github.com/hashicorp/terraform/configs"
 	"go.uber.org/zap"
+)
 
-	"github.com/accurics/terrascan/pkg/iac-providers/output"
+var (
+	skipRulesPattern = regexp.MustCompile(`#ts:skip=\s*(([A-Za-z0-9]+\.?){5})(\s*,\s*([A-Za-z0-9]+\.?){5})*`)
```
williepaul, line 32: Just FYI, I think the rule ID format may go through some changes prior to the next release.

kanchwala-yusuf: Yes, kept that in mind!

pkg/iac-providers/terraform/v12/resource.go
kanchwala-yusuf: Done!

williepaul (approved):
Changes look fine, my only thought is about the skip_rules param when it is null. Shall we mark this in the json as omitempty? It seems like that would simplify some of the test data by not having to include it when null. Also, it prevents an empty field from being printed out in the yaml/json/xml output.

kanchwala-yusuf:
> Changes look fine, my only thought is about the skip_rules param when it is null. Shall we mark this in the json as omitempty? It seems like that would simplify some of the test data by not having to include it when null. Also, it prevents an empty field from being printed out in the yaml/json/xml output.

@williepaul, I agree with you and I do not have a problem using `omitempty` flag as well. The reason it is kept as it because `skip_rules` would be in output of `--config-only` option which is primarily used for debugging. And the thought that keep the `skip_rules` field untouched would help in debugging.

williepaul (approved):

PR #436, opened 2020-12-21
scan command refactor

patilpankaj212:
1. refactor code for scan command
2. more testability for scan command
3. modified existing tests for Run func

### DIFF
--- a/pkg/cli/run.go
+++ b/pkg/cli/run.go
@@ -27,6 +27,7 @@ import (
 	"github.com/accurics/terrascan/pkg/runtime"
 	"github.com/accurics/terrascan/pkg/utils"
 	"github.com/accurics/terrascan/pkg/writer"
+	"github.com/mattn/go-isatty"
 	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/cli/scan.go
```diff
@@ -18,52 +18,15 @@ package cli
...
-	ConfigOnly bool
-
-	// UseColors indicates whether to use color output
-	UseColors bool
-	useColors string // used for flag processing
-
-	// Verbose indicates whether to display all fields in default human readlbe output
-	Verbose bool
-)
+var scanOptions = new(ScanOptions)
```
kanchwala-yusuf, line 29: Also, can have a function like `NewScanOptions()` which returns a new empty `*ScanOptions` struct?

pkg/cli/run.go
```diff
@@ -27,87 +27,185 @@ import (
...
-	iacFilePath, iacDirPath, configFile string, policyPath []string,
-	format, remoteType, remoteURL string, configOnly, useColors, verbose bool) {
+func (s *ScanOptions) Run() error {
 
 	// temp dir to download the remote repo
 	tempDir := filepath.Join(os.TempDir(), utils.GenRandomString(6))
 	defer os.RemoveAll(tempDir)
 
 	// download remote repository
-	path, err := downloadRemoteRepository(remoteType, remoteURL, tempDir)
```
williepaul, line 47: Doesn't seem like downloadRemoteRepository belongs in the scan options class... I think it could be better off separate.

pkg/cli/run.go
```diff
@@ -27,87 +27,185 @@ import (
...
 	// executor output
 	results, err := executor.Execute()
 	if err != nil {
-		return
+		return err
 	}
 
 	// write results to console
-	err = writeResults(results, useColors, verbose, configOnly, format)
+	err = s.writeResults(results)
```
williepaul, line 172: Same comment for writeResults(), it feels weird to add as part of ScanOptions... Perhaps instead, we should have a separate Scanner struct that accepts scan options and performs these tasks... The Scanner object overall could be the thing that orchestrates operations for scan options, policies, normalized input, the scan engine, and results output

williepaul: I'll add that my own previous refactor took the policy input directly into the policy engine, but something felt missing there as well. But having an overall Scanner class that handles this looks like a good option.

pkg/cli/run.go
```diff
@@ -27,87 +27,185 @@ import (
...
+	"github.com/mattn/go-isatty"
 	"go.uber.org/zap"
 )
 
 const (
 	humanOutputFormat = "human"
 )
 
+// ScanOptions represents scan command and its optional flags
+type ScanOptions struct {
```
williepaul, line 39: Thanks, long overdue.

pkg/cli/scan.go
patilpankaj212: I did not add any constructor function for ScanOptions struct because we are not doing any initialisation based on parameters.

pkg/cli/run.go
williepaul: Had a discussion about this w/Pankaj/Yusuf; I meant for the main Scanner object to be within the Executor(), not outside, so since this change is all outside of the Executor, it's not applicable, please disregard.

pkg/cli/run.go
williepaul: n/a

kanchwala-yusuf (approved):

PR #441, opened 2020-12-23
implement scan and skip rules

patilpankaj212:
1. command line options to scan and skip specific rules
2. config file based rule scanning and skipping
3. unit tests


pkg/http-server/file-scan.go
```diff
@@ -86,10 +86,10 @@ func (g *APIHandler) scanFile(w http.ResponseWriter, r *http.Request) {
 	var executor *runtime.Executor
 	if g.test {
 		executor, err = runtime.NewExecutor(iacType, iacVersion, cloudType,
-			tempFile.Name(), "", "", []string{"./testdata/testpolicies"})
+			tempFile.Name(), "", "", []string{"./testdata/testpolicies"}, []string{}, []string{})
```
kanchwala-yusuf, line 89: We should be taking rules to scan and skip for the API server as well. This functionality applies to the scanning for API server as well.

pkg/http-server/remote-repo.go
```diff
@@ -112,7 +112,7 @@ func (s *scanRemoteRepoReq) ScanRemoteRepo(iacType, iacVersion string, cloudType
 
 	// create a new runtime executor for scanning the remote repo
 	executor, err := runtime.NewExecutor(iacType, iacVersion, cloudType,
-		"", iacDirPath, "", policyPath)
+		"", iacDirPath, "", policyPath, []string{}, []string{})
```
kanchwala-yusuf, line 115: We should be taking rules to scan and skip for the API server as well. This functionality applies to the scanning for API server as well.

pkg/policy/opa/engine.go
```diff
@@ -46,13 +46,13 @@ var (
 )
 
 // NewEngine returns a new OPA policy engine
-func NewEngine(policyPath string) (*Engine, error) {
+func NewEngine(policyPath string, scanRules, skipRules []string) (*Engine, error) {
```
kanchwala-yusuf, line 49: Wondering if `scanRules` and `skipRules` can become part of the `Engine{}` struct. 

Would like an opinion from @williepaul on this.

pkg/policy/opa/engine.go
```diff
@@ -249,14 +249,20 @@ func (e *Engine) CompileRegoFiles() error {
...
+func (e *Engine) Init(policyPath string, scanRules, skipRules []string) error {
 	e.context = context.Background()
 
 	if err := e.LoadRegoFiles(policyPath); err != nil {
 		zap.S().Error("error loading rego files", zap.String("policy path", policyPath), zap.Error(err))
 		return err
 	}
 
+	// before compiling the rego files, filter the rules based on scan and skip rules supplied
+	filterRules(e, policyPath, scanRules, skipRules)
```
kanchwala-yusuf, line 261: Would it make sense to have `filterRules` to be tied to the `Engine` struct?

pkg/policy/opa/engine.go
```diff
@@ -249,14 +249,20 @@ func (e *Engine) CompileRegoFiles() error {
...
+
+	// update the rule count
+	e.stats.ruleCount = len(e.regoDataMap)
```
kanchwala-yusuf, line 264: May be it's a good idea to error out if the `ruleCount` is less than 1.

pkg/policy/opa/engine.go
```diff
@@ -395,3 +401,49 @@ func (e *Engine) Evaluate(engineInput policy.EngineInput) (policy.EngineOutput,
...
+		filterScanRules(e, policyPath, scanRules)
+	}
+
+	// apply skip rules
+	if len(skipRules) > 0 {
+		filterSkipRules(e, policyPath, skipRules)
+	}
+}
+
+func filterScanRules(e *Engine, policyPath string, scanRules []string) {
```
kanchwala-yusuf, line 417: Can we tie `filterScanRules` with the `Engine{}` struct?

pkg/policy/opa/engine.go
```diff
@@ -395,3 +401,49 @@ func (e *Engine) Evaluate(engineInput policy.EngineInput) (policy.EngineOutput,
...
+	}
+	if len(tempMap) == 0 {
+		zap.S().Warnf("scan rule id's: %+v not found in policy path: %s", scanRules, policyPath)
+	}
+
+	// the regoDataMap should only contain regoData for supplied scan rules
+	e.regoDataMap = tempMap
+}
+
+func filterSkipRules(e *Engine, policyPath string, skipRules []string) {
```
kanchwala-yusuf, line 438: Can we tie `filterSkipRules` with the `Engine{}` struct?

pkg/policy/opa/engine_test.go
```diff
@@ -1 +1,98 @@
...
+					regoDataMap: getTestRegoDataMap(10),
+				},
+				policyPath: testPolicyPath,
+				scanRules:  []string{"Rule.11", "Rule.12", "Rule.13"},
+			},
+			assert:      true,
+			regoMapSize: 0,
+		},
+		{
+			name: "skip rules test",
```
kanchwala-yusuf, line 54: Can we have a test for `scan and skip rules present`?

pkg/runtime/executor.go
```diff
@@ -17,12 +17,27 @@
...
 
+	"github.com/accurics/terrascan/pkg/config"
 	iacProvider "github.com/accurics/terrascan/pkg/iac-providers"
 	"github.com/accurics/terrascan/pkg/notifications"
 	"github.com/accurics/terrascan/pkg/policy"
 	opa "github.com/accurics/terrascan/pkg/policy/opa"
+	"github.com/pelletier/go-toml"
+)
+
+const (
```
kanchwala-yusuf, line 32: I would suggest creating a new file `rules.go` and move these constants in that file.

pkg/runtime/executor.go
```diff
@@ -17,12 +17,27 @@
...
+	rulesKey     = "rules"
+	scanRulesKey = "scan-rules"
+	skipRulesKey = "skip-rules"
+)
+
+var (
```
kanchwala-yusuf, line 38: Another candidate for `rules.go`

pkg/runtime/executor.go
```diff
@@ -139,3 +168,68 @@ func (e *Executor) Execute() (results Output, err error) {
 	// successful
 	return results, nil
 }
+
+// read the config file and update scan and skip rules
+func (e *Executor) initScanAndSkipRules() error {
```
kanchwala-yusuf, line 173: Another good candidate for `rules.go`

pkg/runtime/executor.go
```diff
@@ -139,3 +168,68 @@ func (e *Executor) Execute() (results Output, err error) {
...
+			// read skip rules in the toml tree
+			if err := initRules(e, data, skipRulesKey); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func initRules(e *Executor, tree *toml.Tree, key string) error {
```
kanchwala-yusuf, line 199: Another good candidate for `rules.go`

pkg/runtime/executor.go
```diff
@@ -139,3 +168,68 @@ func (e *Executor) Execute() (results Output, err error) {
...
+		} else {
+			e.skipRules = append(e.skipRules, rules...)
+		}
+	} else {
+		zap.S().Debugf("key '%s' not found in the config file: %s", key, e.configFile)
+	}
+	return nil
+}
+
+func getRulesInTomlTree(tree *toml.Tree, configFile, key string) ([]string, error) {
```
kanchwala-yusuf, line 217: Another good candidate for `rules.go`.

But, I feel this should function should be a part of the `config` package. This function can be a helper for returning rules from the `config` package.

pkg/runtime/executor.go
```diff
@@ -139,3 +168,68 @@ func (e *Executor) Execute() (results Output, err error) {
...
+
+func initRules(e *Executor, tree *toml.Tree, key string) error {
+	rules, err := getRulesInTomlTree(tree, e.configFile, key)
+	if err != nil {
+		zap.S().Error("error reading config file", zap.Error(err))
+		return err
+	}
+	if len(rules) > 0 {
+		if key == scanRulesKey {
+			e.scanRules = append(e.scanRules, rules...)
```
kanchwala-yusuf, line 207: Not the right place to append to scan rules, this function should just return the list of rules. Caller function should take responsibility of using the list of rules. In this way this function can become more generic and we would not require to pass the `Executor` object to this.

pkg/policy/interface.go
```diff
@@ -18,7 +18,8 @@ package policy
 
 // Engine Policy Engine interface
 type Engine interface {
-	Init(string) error
+	// to initialize engine with policy path, scan and skip rules
```
kanchwala-yusuf, line 21: Comment should start with `Init`

pkg/runtime/executor.go
kanchwala-yusuf, line 173: Also, `initRules()` as a function name should suffice, isn't it?

pkg/runtime/executor.go
```diff
@@ -139,3 +168,68 @@ func (e *Executor) Execute() (results Output, err error) {
 	// successful
 	return results, nil
 }
+
+// read the config file and update scan and skip rules
+func (e *Executor) initScanAndSkipRules() error {
+	if e.configFile != "" {
```
kanchwala-yusuf, line 174: Would be a better idea to do:
```
if e.configFile == "" {
    return nil
}
```

pkg/runtime/executor.go
```diff
@@ -139,3 +168,68 @@ func (e *Executor) Execute() (results Output, err error) {
...
+		configData, err := config.LoadConfig(e.configFile)
```
kanchwala-yusuf, line 175: It should not be the responsibility of this package to `LoadConfig`, it should happen in `config` package. There should be handlers in the `config` package to return what's required by this package.

kanchwala-yusuf (changes requested):

pkg/http-server/file-scan.go
patilpankaj212: Implemented

pkg/http-server/remote-repo.go
patilpankaj212: Implemented

pkg/policy/opa/engine.go
patilpankaj212: Added a function FilterRules(string, []string, []string) to Engine interface and moved the Init function out of the NewEngine function.

pkg/policy/opa/engine.go
patilpankaj212: implemented as part of interface

pkg/policy/opa/engine.go
patilpankaj212: multiple engines are created based on policy path slice, cannot error out.

pkg/policy/opa/engine_test.go
patilpankaj212: added a test for it.

pkg/runtime/executor.go
patilpankaj212: this was removed as part of config reader refactor

pkg/runtime/executor.go
patilpankaj212: removed in the refactored code

pkg/runtime/executor.go
patilpankaj212: done as part of config reader refactor

pkg/policy/opa/engine.go
williepaul: At this point, adding to an interface makes sense. Later, we should have a top-level "Scanner" struct that manages interactions between the scan engine, filters, policies/rules, and results. Today, that's all considered the "scan engine", but I just want to limit that piece to actual scanning only. Policies, rules, rule filters, and results processing are applicable to different scan engines, so in the future they should work independently of it.

pkg/config/global.go
```diff
@@ -31,87 +30,54 @@ const (
...
+	configReader, err := NewTerrascanConfigReader(configFile)
 	if err != nil {
-		return p, ErrNotPresent
+		return err
 	}
 
-	keyConfig := config.Get(policyConfigKey)
-	if keyConfig == nil {
-		return p, errTomlKeyNotPresent
+	if len(configReader.GetPolicyConfig().BasePath) > 0 {
```
williepaul, line 68: I think we shouldn't do these length checks. The reason is, if the global config happens to contain an existing config, running loadConfigFile could leave the user with a partially loaded config file (values from old config + some values from new config), and that is misleading to the user. I would expect if the new config file has an empty path, an empty path should be loaded. That is an error the user would have to fix. I would just remove the length check and set it directly: Global.Policy.BasePath = configReader.GetPolicyConfig().BasePath, or simply assign the entire config in one shot.

williepaul (changes requested):

kanchwala-yusuf (approved):

williepaul (approved):

PR #443, opened 2020-12-24
bump go to 1.15

chenrui333:
Bump go to the latest (go v1.15)

It is already built and packaged with go 1.15 in the homebrew side https://github.com/Homebrew/homebrew-core/pull/64996


cesar-rodriguez (approved):
Thanks!

PR #444, opened 2020-12-24
deps: add dependabot support

chenrui333:
Add dependabot support which would automate the dependencies management, I am using it pretty extensively, so I feel it is a great addon for this project.


cesar-rodriguez (approved):
This is very helpful. Thanks!

PR #455, opened 2021-01-06
rule skipping for resources in k8s

patilpankaj212:


pkg/iac-providers/kubernetes/v1/normalize.go
```diff
@@ -126,3 +136,27 @@ func (k *K8sV1) Normalize(doc *utils.IacDocument) (*output.ResourceConfig, error
 
 	return &resourceConfig, nil
 }
+
+func readSkipRulesFromAnnotations(annotations map[string]interface{}, resourceID string) []string {
+
+	if _, ok := annotations[terrascanSkipRules]; !ok {
```
williepaul, line 142: Can you remove the assignment on 148 and just specify skipRulesFromAnnotations in place of _ here?

williepaul (approved):

PR #465, opened 2021-01-10
recognize metadata.generateName

acc-jon:
As described at https://kubernetes.io/docs/reference/using-api/api-concepts/#generated-values, the metadata.name may sometimes be generated based on metadata.generateName.  Dynamically generated names are populated with unique values at runtime, meaning that metadata.name might be empty when Terrascan is run.  This change uses the value of metadata.generateName for the name when metadata.name is empty.  This can help, for example, with CRDs that rely on generateName.

### DIFF
--- a/pkg/iac-providers/kubernetes/v1/normalize.go
+++ b/pkg/iac-providers/kubernetes/v1/normalize.go
@@ -38,9 +38,18 @@ var (
 
 // k8sMetadata is used to pull the name, namespace types and annotations for a given resource
 type k8sMetadata struct {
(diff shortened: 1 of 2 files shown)


jlk:
The duplicate code doesn't look like an issue to me - it looks like it's mainly the expected result from unit tests. I guess you could try to define that as a struct and then change the bits for each test, but that seems like over-engineering...

williepaul:
SonarCloud has a different mode for analyzing source vs test files, will check how to configure that.

williepaul (approved):

PR #478, opened 2021-01-14
Policy update 2021 01 14

williepaul:


acc-jon (commented):
I didn't review the rego in detail, but this seems to include the policies I need for my blog post.  I just need to verify two more:
- Container should not be privileged
- Minimize admission of containers with capabilities assigned

PR #501, opened 2021-01-19
feature: add options to specify desired severity level of violations to be reported

### LINKED ISSUE #392: Add --severity options
I've explore `terrascan` tools to filter an output by severity level but can't found it, it will be good if the tools can filter an output by risks and possible to embeded into CI/CD pipeline.

dev-gaur:
fixes #392

### DIFF
--- a/pkg/cli/scan.go
+++ b/pkg/cli/scan.go
@@ -63,5 +63,6 @@ func init() {
 	scanCmd.Flags().BoolVarP(&scanOptions.Verbose, "verbose", "v", false, "will show violations with details (applicable for default output)")
 	scanCmd.Flags().StringSliceVarP(&scanOptions.scanRules, "scan-rules", "", []string{}, "one or more rules to scan (example: --scan-rules=\"ruleID1,ruleID2\")")
 	scanCmd.Flags().StringSliceVarP(&scanOptions.skipRules, "skip-rules", "", []string{}, "one or more rules to skip while scanning (example: --skip-rules=\"ruleID1,ruleID2\")")
+	scanCmd.Flags().StringVar(&scanOptions.severity, "severity", "", "minimum severity level of the policy violations to be reported by terrascan")
 	RegisterCommand(rootCmd, scanCmd)
 }
--- a/pkg/http-server/file-scan.go
+++ b/pkg/http-server/file-scan.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 
 	"github.com/accurics/terrascan/pkg/runtime"
+	"github.com/accurics/terrascan/pkg/utils"
 	"github.com/gorilla/mux"
 	"go.uber.org/zap"
 )
@@ -89,6 +90,9 @@ func (g *APIHandler) scanFile(w http.ResponseWriter, r *http.Request) {
 	scanRulesValue := r.FormValue("scan_rules")
 	skipRulesValue := r.FormValue("skip_rules")
 
+	// severity is the minimum severity level of violations that the user want to get informed about: low, medium or high
+	severity := r.FormValue("severity")
+
 	if scanRulesValue != "" {
 		scanRules = strings.Split(scanRulesValue, ",")
 	}
@@ -97,14 +101,18 @@ func (g *APIHandler) scanFile(w http.ResponseWriter, r *http.Request) {
 		skipRules = strings.Split(skipRulesValue, ",")
 	}
 
+	if severity != "" {
+		severity = utils.EnsureUpperCaseTrimmed(severity)
+	}
+
 	// create a new runtime executor for scanning the uploaded file
 	var executor *runtime.Executor
 	if g.test {
 		executor, err = runtime.NewExecutor(iacType, iacVersion, cloudType,
-			tempFile.Name(), "", "", []string{"./testdata/testpolicies"}, scanRules, skipRules)
+			tempFile.Name(), "", "", []string{"./testdata/testpolicies"}, scanRules, skipRules, severity)
 	} else {
 		executor, err = runtime.NewExecutor(iacType, iacVersion, cloudType,
-			tempFile.Name(), "", "", []string{}, scanRules, skipRules)
+			tempFile.Name(), "", "", []string{}, scanRules, skipRules, severity)
 	}
 	if err != nil {
 		zap.S().Error(err)
--- a/pkg/http-server/file-scan_test.go
+++ b/pkg/http-server/file-scan_test.go
@@ -31,6 +31,7 @@ func TestUpload(t *testing.T) {
 		cloudType  string
 		scanRules  []string
 		skipRules  []string
+		severity   string
 		wantStatus int
 	}{
 		{
@@ -56,6 +57,15 @@ func TestUpload(t *testing.T) {
 			iacType:    testIacType,
 			cloudType:  testCloudType,
 			wantStatus: http.StatusOK,
+		},
+		{
+			name:       "valid file scan default iac version, with invalid severity level input",
+			path:       testFilePath,
+			param:      testParamName,
+			iacType:    testIacType,
+			severity:   "HGIH",
+			cloudType:  testCloudType,
+			wantStatus: http.StatusBadRequest,
 		},
 		{
 			name:       "invalid iacType",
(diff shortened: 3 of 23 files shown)


pkg/http-server/file-scan.go
```diff
@@ -37,6 +37,7 @@ func (g *APIHandler) scanFile(w http.ResponseWriter, r *http.Request) {
 	var (
 		iacType    = params["iac"]
 		iacVersion = params["iacVersion"]
+		severity   = params["severity"]
```
kanchwala-yusuf, line 40: `severity` would not be received in the url params

dev-gaur: then how?

dev-gaur: @kanchwala-yusuf changes made!

pkg/policy/opa/engine.go
```diff
@@ -420,7 +420,7 @@ func (e *Engine) Evaluate(engineInput policy.EngineInput) (policy.EngineOutput,
 }
 
 // FilterRules will apply the scan and skip rules
```
patilpankaj212, line 422: updating comment makes sense here too?

pkg/policy/opa/engine.go
```diff
@@ -465,3 +469,21 @@ func (e *Engine) filterSkipRules(policyPath string, skipRules []string) {
...
+	// temporary map to store data from original rego data map
+	tempMap := make(map[string]*RegoData)
+	for ruleID, regoData := range e.regoDataMap {
+
+		if utils.CheckSeverity(regoData.Metadata.Severity, severity) {
+			tempMap[ruleID] = regoData
+		}
+	}
+	if len(tempMap) == 0 {
+		zap.S().Warnf("policy path: %s, doesn't have any rule matching the severity level : %s", policyPath, severity)
```
patilpankaj212, line 484: this log should not be at warn level, nothing bad has happened.

pkg/runtime/validate.go
```diff
@@ -102,6 +103,11 @@ func (e *Executor) ValidateInputs() error {
 	}
 	zap.S().Debugf("using policy path %v", e.policyPath)
 
+	if len(e.severity) > 0 && !utils.ValidateSeverityInput(e.severity) {
+		return errSeverityNotSupported
```
patilpankaj212, line 107: should we error out when incorrect severity is passed by user?
I think we should only log regarding it and proceed with the scan.

pkg/http-server/file-scan_test.go
```diff
@@ -57,6 +58,15 @@ func TestUpload(t *testing.T) {
 			cloudType:  testCloudType,
 			wantStatus: http.StatusOK,
 		},
+		{
+			name:       "valid file scan default iac version with invalide severity type input",
```
patilpankaj212, line 62: typo?

patilpankaj212 (changes requested):

pkg/policy/opa/engine.go
dev-gaur: I did it following the filterSkipRules and filterScanRules methods. Doesn't hurt to inform the user :D

pkg/runtime/validate.go
dev-gaur: IMO, we should explicitly inform the user about their human error. A user might enter a invalid value "hihg" expecting only HIGH severity violations to be reported. But if assume the defaults and proceed with scan in this case, they'll get reported the violations of all the severity levels, opposed to the user's expectation. This erroring out way will help the user make robust decisions while creating CI / CD pipelines as well.

WDYT @williepaul @kanchwala-yusuf @jlk @acc-jon @amirbenv

pkg/policy/opa/engine.go
patilpankaj212: yeah, informing or warning is the difference :D

pkg/runtime/validate.go
patilpankaj212: IMO, the scan command (or any other command) should error out only when an option value is critical for the scan command to proceed eg: the file/directory passed for scan.

amirbenv: I think if a user specifies a severity, and inputs an error, we should error out and tell them they inputted an unsupported severity. Otherwise, the scan will not run as they expect, and they could either not notice, or notice and think terrascan has a bug.

acc-jon: I think we should treat it similar to other "picklist" options like -i: there are specific valid values for the option, and invalid options should trigger a usage error.

pkg/runtime/rules.go
```diff
@@ -26,5 +26,10 @@ func (e *Executor) initRules() error {
 	if len(configReader.GetRules().SkipRules) > 0 {
 		e.skipRules = append(e.skipRules, configReader.GetRules().SkipRules...)
 	}
+
+	// specify severity of violations to be reported
+	if len(configReader.GetSeverity().Level) > 0 {
```
patilpankaj212, line 31: Does this introduces a scope for a bug related to validation of severity level? Since the validation happens before init is called, e.severity will be assigned with a value in the config file and would not be validated.

pkg/cli/scan.go
```diff
@@ -63,5 +63,6 @@ func init() {
 	scanCmd.Flags().BoolVarP(&scanOptions.Verbose, "verbose", "v", false, "will show violations with details (applicable for default output)")
 	scanCmd.Flags().StringSliceVarP(&scanOptions.scanRules, "scan-rules", "", []string{}, "one or more rules to scan (example: --scan-rules=\"ruleID1,ruleID2\")")
 	scanCmd.Flags().StringSliceVarP(&scanOptions.skipRules, "skip-rules", "", []string{}, "one or more rules to skip while scanning (example: --skip-rules=\"ruleID1,ruleID2\")")
+	scanCmd.Flags().StringVar(&scanOptions.severity, "severity", "", "minimum severity level of the policy violations that you want to be reported")
```
kanchwala-yusuf, line 66: Let's confirm, if we want the help to in second person. Usually it goes like:
```
minimum severity level of the policy violations to be reported by terrascan
```

pkg/policy/opa/engine.go
```diff
@@ -465,3 +469,21 @@ func (e *Engine) filterSkipRules(policyPath string, skipRules []string) {
 		}
 	}
 }
+
+func (e *Engine) filterBySeverity(policyPath string, severity string) {
```
kanchwala-yusuf, line 473: we can have a common declaration for input vars some like:
```
func (e *Engine) filterBySeverity(policyPath, severity string) {

```

pkg/policy/opa/engine.go
kanchwala-yusuf: agree with @patilpankaj212, do think there is a need for a warning, may be a `debug` level log

pkg/runtime/executor.go
```diff
@@ -27,22 +27,23 @@ import (
...
+	skipRules     []string
+	iacProvider   iacProvider.IacProvider
+	policyEngines []policy.Engine
+	notifiers     []notifications.Notifier
+	severity      string
 }
 
 // NewExecutor creates a runtime object
-func NewExecutor(iacType, iacVersion string, cloudType []string, filePath, dirPath, configFile string, policyPath, scanRules, skipRules []string) (e *Executor, err error) {
+func NewExecutor(iacType, iacVersion string, cloudType []string, filePath, dirPath, configFile string, policyPath, scanRules, skipRules []string, severity string) (e *Executor, err error) {
```
kanchwala-yusuf, line 46: The function input vars are increasing, we may need to do something about this. Not necessarily in this PR :)

pkg/runtime/rules.go
kanchwala-yusuf: I think, there is a valid point. Here we are assigning the value read from the config file, which will override the value set on the CLI.

kanchwala-yusuf (changes requested):

pkg/runtime/rules.go
dev-gaur: @patilpankaj212 @kanchwala-yusuf  PTAL at the executor.go function call flow. This has been resolved there. 

https://github.com/accurics/terrascan/blob/8373b84d1df6dfb84fe7c8c756e1d1dbdf983937/pkg/runtime/executor.go#L47

patilpankaj212: got it. skip rules, scan rules and severity is update with config file values before the executor init is called.

patilpankaj212 (dismissed):
LGTM

pkg/runtime/executor.go
```diff
@@ -51,8 +52,24 @@ func NewExecutor(iacType, iacVersion string, cloudType []string, filePath, dirPa
 		iacType:    iacType,
 		iacVersion: iacVersion,
 		configFile: configFile,
-		scanRules:  scanRules,
-		skipRules:  skipRules,
+	}
+
+	// read config file and update scan and skip rules
+	if err := e.initRuleSetFromConfigFile(); err != nil {
```
patilpankaj212, line 58: we should add tests for the new behaviour that we added i.e. command line values overrides config file values.

patilpankaj212 (changes requested):

patilpankaj212 (approved):
LGTM !!

kanchwala-yusuf (approved):

PR #531, opened 2021-02-03
refactor init command for robust policy download checks

dev-gaur:
Currently we delete the existing locally cached policy set before making ensuring that the new policy set can be download ( there can be absence of internet connection or maybe github servers down leading to disability in downloading new policy set).

This PR will refactor the init process so that we download the new policy set before sweeping off the existing locally cached policy set.


jlk (approved):
LGTM

Issue #551, opened 2021-02-16
Could not get terrascan init to work - would not download policy documents


swails:
* terrascan version:  1.3.2
* Operating System: Linux x86-64

### Description

I could not initialize terrascan after first downloading it

### What I Did

```bash
$ wget https://github.com/accurics/terrascan/releases/download/v1.3.2/terrascan_1.3.2_Linux_x86_64.tar.gz
$ tar zxvf terrascan_1.3.2_Linux_x86_64.tar terrascan
$ mv terrascan /usr/local/bin
$ terrascan init -l debug
2021-02-16T14:50:55.754-0500	debug	config/global.go:59	global config env variable is not specified
2021-02-16T14:50:55.754-0500	debug	initialize/run.go:41	initializing terrascan
2021-02-16T14:50:55.755-0500	error	initialize/run.go:69	failed to download policies. error: 'repository already exists'
2021-02-16T14:50:55.755-0500	error	cli/init.go:42	failed to initialize terrascan
```

I couldn't find any documentation that gave me an ideas of what to try next...

dev-gaur:
Hello @swails , could you look for a terrascan folder inside the /tmp path or $TMPDIR  path ? When found, please detete that terrascan directory and try again? Please report the result back here.
 
It seems the repository was cloned in the tmp location from earlier runs and wasn't removed by the code logic.

swails:
No dice:

```bash
$ rm -fr /tmp/terrascan/
$ terrascan init -l debug
2021-02-16T16:04:45.284-0500	debug	config/global.go:59	global config env variable is not specified
2021-02-16T16:04:45.284-0500	debug	initialize/run.go:41	initializing terrascan
2021-02-16T16:04:47.868-0500	error	cli/init.go:42	failed to initialize terrascan
```

Although the "repository already exists" error is gone...

dev-gaur:
Aah, I think the failure is happening at this line now 
 https://github.com/accurics/terrascan/blob/v1.3.2/pkg/initialize/run.go#L101 
 
 this os.Rename( ) method is the golang equivalent of a unix `mv` command.  There might be some trouble with the file permissions.

dev-gaur:
your working directory is probably write protected

dev-gaur:
@swails did it work after changing file permissions?

dev-gaur:
Closing this issue for now as v1.3.3 has been release with a fix. Please re-open it if you still face this issue.

PR #552, opened 2021-02-16
fix the 'repo already exist' bug and improve error logging for terrascan init

dev-gaur:
Fixes a part of  #551

### DIFF
--- a/pkg/initialize/run.go
+++ b/pkg/initialize/run.go
@@ -18,8 +18,8 @@ package initialize
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
-	"path/filepath"
 
 	"github.com/accurics/terrascan/pkg/config"
 	"go.uber.org/zap"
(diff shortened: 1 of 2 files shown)


pkg/initialize/run.go
```diff
@@ -61,6 +61,12 @@ func DownloadPolicies() error {
 
 	tempPath := filepath.Join(os.TempDir(), "terrascan")
```
kanchwala-yusuf, line 62: Hey @dev-gaur,
Would it make sense to use `ioutil.TempDir()` instead of `os.TempDir()`

kanchwala-yusuf (changes requested):

pkg/initialize/run.go
dev-gaur: eureka!

dev-gaur: changes made. PTAL again

kanchwala-yusuf (approved):

PR #554, opened 2021-02-17
Update README.md in build your own docker section

versus:
Update build your own docker section in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,17 +129,33 @@ $ ./bin/terrascan
 ```
 
 ### To build your own docker, refer to this example (Alpine Linux):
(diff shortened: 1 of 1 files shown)


kanchwala-yusuf:
Hey @versus,
Thanks for making this contribution. But we do have a `make docker-build` command which would directly create a docker image for a user.

Do you think, this should suffice? May be we can document the `make docker-build` command

versus:
I don't mind

dev-gaur:
Hi @versus , I've acommodated your commits into an updated pr #555 . PTAL and verify the instructions if they work for you. closing this.

PR #555, opened 2021-02-17
Update docker build and push instructions in README 

dev-gaur:


scripts/docker-build.sh
```diff
@@ -4,8 +4,25 @@ set -o errexit
...
 set -o pipefail
 
+if [ $# -eq 2 ]; then
+    DOCKER_REPO=$1
+    TAG=$2
+fi
+
+if [ $# -eq 1 ]; then
+    DOCKER_REPO=$1
+fi
```
kanchwala-yusuf, line 14: I am just wondering, rather than doing all of this, would it make sense for the user to use the existing `make docker-build` and then use `docker tag` for creating an image in user's docker registry.

If we accept variables from the CLI, we should ideally validate it and I feel that may not be necessary if we document the `docker tag` command.

What do you think @dev-gaur?

dev-gaur: an image tag being used locally, should be user's responsibility and i don't think any prod-level image usage will occur locally.
For validation, docker can validate the username & repo name and tag for the respective user while pushing to registry.

dev-gaur: Also, if the user simply runs `make docker-build` without any arguments, then simply the image will be tagged as `accurics/terrascan:<git-commit>` , just like before.

dev-gaur: I feel keeping the make target will be better as we'll be able to edit the scripts underneath in the future as well without breaking the usage command everytime.

Issue #569, opened 2021-02-26
accurics.gcp.NS.109 too restrictive?


b4nst:
`accurics.gcp.NS.109` ensure GKE Control Plane is not public and yells HIGH severity without taking account of whether or not the `master_authorized_networks_config` is set.

 The Google [recommended way](https://cloud.google.com/kubernetes-engine/docs/how-to/private-clusters) of giving access to an external resources is by using `master-authorized-networks`.

It seems to me that if we use `master_authorized_networks_config` we should get a `MEDIUM` and no `HIGH`.  That would imply adding a rule for `master_authorized_networks_config` and taking into account `master_authorized_networks_config` for `accurics.gcp.NS.109`

harkirat22:
@b4nst this makes total sense. Thank you for raising the issue. We will resolve this soon.

PR #575, opened 2021-03-02
fix:adding new policy and enhancing the noisy policy

harkirat22:
fix proposed for #569 includes:
1. Enhancing the existing policy to only throw violation if private_endpoint is not true and master_authorized_networks_config is not used.
2. Included new policy with medium violation provided master_authorized_networks_config CIDR block has a public IP address associated.

### DIFF
--- a/pkg/policies/opa/rego/gcp/google_container_cluster/AC-GC-IS-CC-M-0367.json
+++ b/pkg/policies/opa/rego/gcp/google_container_cluster/AC-GC-IS-CC-M-0367.json
@@ -1,0 +1,10 @@
+{
+    "name": "gkeControlPlanePublicAccess",
+    "file": "gkeControlPlanePublicAccess.rego",
+    "template_args": null,
+    "severity": "Medium",
+    "description": "GKE Control Plane is exposed to few public IP addresses using master-authorized-network-config",
+    "reference_id": "AC-GC-IS-CC-M-0367",
+    "category": "Infrastructure Security",
+    "version": 1
+}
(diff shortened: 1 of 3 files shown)


pkg/policies/opa/rego/gcp/google_container_cluster/AC-GC-IS-CC-M-0367.json
```diff
@@ -0,0 +1,10 @@
+{
+    "name": "gkeControlPlanePublicAccess",
+    "file": "gkeControlPlanePublicAccess.rego",
+    "template_args": null,
+    "severity": "Medium",
+    "description": "GKE Control Plane is exposed to few public IP addresses using master-authorized-network-config",
+    "reference_id": "AC-GC-IS-CC-M-0367",
```
gaurav-gogia, line 7: Can we use the new rule ref id format here? :)

gaurav-gogia (changes requested):

pkg/policies/opa/rego/gcp/google_container_cluster/AC-GC-IS-CC-M-0367.json
dev-gaur: @harkirat22

harkirat22: @gaurav-gogia . I would prefer to make the rule-reference id update only once its formally documented as our previous updates did break some of the end user's pipelines.

gaurav-gogia: @harkirat22 I see, sounds good

gaurav-gogia (approved):

PR #579, opened 2021-03-03
adds support to use 'config_only' attribute in api server's file scan

patilpankaj212:
1. config_only value for api server file scan
2. unit tests

### DIFF
--- a/pkg/http-server/file-scan.go
+++ b/pkg/http-server/file-scan.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"strconv"
 	"strings"
 
 	"github.com/accurics/terrascan/pkg/runtime"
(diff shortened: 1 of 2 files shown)


pkg/http-server/file-scan.go
```diff
@@ -93,6 +95,18 @@ func (g *APIHandler) scanFile(w http.ResponseWriter, r *http.Request) {
 	// severity is the minimum severity level of violations that the user want to get informed about: low, medium or high
 	severity := r.FormValue("severity")
 
+	// read config_only from the form data
+	configOnlyValue := r.FormValue("config_only")
+	if configOnlyValue != "" {
+		configOnly, err = strconv.ParseBool(configOnlyValue)
+		if err != nil {
+			errMsg := fmt.Sprintf("error while reading 'config_only' value. error: '%v'", err)
```
kanchwala-yusuf, line 103: May be we can return the incorrect value received as part of the request, so that the caller understands his mistake.

kanchwala-yusuf (changes requested):

pkg/http-server/file-scan.go
patilpankaj212: Yes, that is part of the err.

kanchwala-yusuf: Ok, makes sense

kanchwala-yusuf (approved):

PR #589, opened 2021-03-06
update helm default chart name and namespace values

williepaul:
- this changes the helm chart default name/namespace values to match a default deployment of a given helm chart

### DIFF
--- a/pkg/iac-providers/helm/v3/load-dir.go
+++ b/pkg/iac-providers/helm/v3/load-dir.go
@@ -149,6 +149,8 @@ func (h *HelmV3) createHelmChartResource(chartPath string, chartData map[string]
 // renderChart renders a helm chart with the given template files and values
 // returns and IaC document for each rendered file
 func (h *HelmV3) renderChart(chartPath string, chartMap helmChartData, templateDir string, templateFiles []*string, valueMap map[string]interface{}) ([]*utils.IacDocument, error) {
+	const defaultNamespaceName = "default"
+
 	iacDocuments := make([]*utils.IacDocument, 0)
 	logger := zap.S().With("helm chart path", chartPath)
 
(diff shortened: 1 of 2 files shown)


pkg/iac-providers/helm/v3/load-dir.go
```diff
@@ -195,8 +197,8 @@ func (h *HelmV3) renderChart(chartPath string, chartMap helmChartData, templateD
 
 	// create rendered values
 	options := chartutil.ReleaseOptions{
-		Name:      defaultChartName,
-		Namespace: chartName + "-namespace",
+		Name:      chartName,
+		Namespace: defaultNamespaceName,
```
kanchwala-yusuf, line 201: Would `Namespace` always be `default`?  Why did we previously have this value equal to `chartName + "-namespace"`? I am just trying to understand how is this change affecting terrascan? and what are the repercussions of not having this.

williepaul: Good question. When first adding Helm support, I was not sure where the name field was being pulled from. I remember wanting to use the original chart name, but from the usage in some of their unit tests, it looked arbitrary. While working on some drift mapping fixes, I noticed Helm really was using the Chart name in this field, and hence the name change.

As for the namespace change, "default" is just used by default. However, as you might have wondered, it may be set to something else. Helm takes a namespace command as input, and that namespace specified on the command line is what is fed into this value. We don't currently have a way of passing variables into our provider implementations, but I think we should add some option to do so, for example -D<key>=<value>, where in this case a user could specify something like -Dnamespace=mynamespace. It's important to us, since without a way to do this, we will not be able to map resources in custom namespaces correctly.

kanchwala-yusuf: @williepaul , I am wondering, is PR partially fixing the problem? And the complete solution would be to accept options from the user and to exact mappings?

jlk: What Willie and I were trying to figure out when looking at this last week: `helm` by default will use the namespace of the current kubernetes context. So how do we map that into a static scan?

williepaul: @kanchwala-yusuf that is correct. A complete solution should take in command line input with data to replace the chart name and namespace as needed. However, these values are suitable defaults, and are currently needed and used by the k8s scan and drift mapping...

kanchwala-yusuf (approved):

PR #593, opened 2021-03-09
Server mode: take file extension from uploaded file

jlk:
Changes to take temp file extension from uploaded file so we can support more than .tf when in server mode.

Alternately - I was thinking of adding GetDefaultExtension() to each of the iac-providers and then naming the temp file based on the iac type passed. That might be a little more secure, but not by very much. Gimme thoughts...

### DIFF
--- a/pkg/http-server/file-scan.go
+++ b/pkg/http-server/file-scan.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"path"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


patilpankaj212 (approved):
LGTM!!

kanchwala-yusuf:
>  I was thinking of adding GetDefaultExtension() to each of the iac-providers and then naming the temp file based on the iac type passed.

It is good an idea to add extensions based on iac-type. But, the changes in the current PR look good to me. We can add extensions based on iac-types if the current implementation fails for some scenario.

kanchwala-yusuf (approved):

PR #605, opened 2021-03-10
Category flag e2e tests

gaurav-gogia:


patilpankaj212 (approved):
LGTM!!
We should later add a test to assert on the output too.

Issue #611, opened 2021-03-10
Potential nil-dereference found while fuzzing


jlk:
* terrascan version:

### Description

While testing fuzzbuzz, an issue was discovered where we attempt to use a yaml decoder without testing if it is non-nil

In `pkg/utils/yaml.go` around line 103:
```
	dec := yaml.NewDecoder(bytes.NewReader(byteArray))

	i := 0
	for {
		// each iteration extracts and marshals one yaml document
		var value interface{}
		err := dec.Decode(&value)
```

PR #612, opened 2021-03-10
Check for nil decoder before use

jlk:
simple add of if test before using decoder.

### DIFF
--- a/pkg/utils/yaml.go
+++ b/pkg/utils/yaml.go
@@ -103,7 +103,7 @@ func ScanIacDocumentsFromYaml(scanner *bufio.Scanner, byteArray []byte, filePath
 	dec := yaml.NewDecoder(bytes.NewReader(byteArray))
 
 	if dec == nil {
-		err := fmt.Errorf("Error cretaing yaml decoder in ScanIacDocumentsFromYaml()")
+		err := fmt.Errorf("error cretaing yaml decoder in ScanIacDocumentsFromYaml()")
 		return iacDocumentList, err
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/utils/yaml.go
```diff
@@ -102,6 +102,11 @@ func ScanIacDocumentsFromYaml(scanner *bufio.Scanner, byteArray []byte, filePath
 
 	dec := yaml.NewDecoder(bytes.NewReader(byteArray))
 
+	if dec == nil {
```
patilpankaj212, line 105: Though the nil check is good, looking at the yaml.NewDecoder method, it will either panic or return a non nil decoder.

patilpankaj212 (commented):
added review.

dev-gaur:
I think the CI will pass here even if you don't include the changes in go.mod and go.sum . I understand that running `make validate` locally must have required a `go mod vendor` first.

pkg/utils/yaml.go
```diff
@@ -102,6 +102,11 @@ func ScanIacDocumentsFromYaml(scanner *bufio.Scanner, byteArray []byte, filePath
...
+		err := fmt.Errorf("error cretaing yaml decoder in ScanIacDocumentsFromYaml()")
```
dev-gaur, line 106: hey @jlk , I think the user wouldn't care to know the function name here. so we can keep the error message generic and simple. also, creating is mispelled

dev-gaur: requsting this change

dev-gaur (changes requested):

jlk:
Sorry for ignoring this. Upon looking closer, the issue was in the library we're using, so will open a PR with them. Closing this PR.

dev-gaur:
#611

PR #626, opened 2021-03-19
Update route53LoggingDisabled.rego to ignore private zones

matt-slalom:
The AWS module route53LoggingDisabled erroneously requires CloudWatch logging for a private Route 53 zone. Per the documentation, query logging only works when the Route53 hosted zone is public.

See https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route53_query_log
and
https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/query-logs.html?console_help=true#query-logs-configuring

### DIFF
--- a/pkg/policies/opa/rego/aws/aws_route53_query_log/route53LoggingDisabled.rego
+++ b/pkg/policies/opa/rego/aws/aws_route53_query_log/route53LoggingDisabled.rego
@@ -2,7 +2,20 @@ package accurics
 
 {{.prefix}}route53LoggingDisabled[route.id] {
   route := input.aws_route53_zone[_]
+  vpc := route.config.dynamic[_].vpc
   not input.aws_route53_query_log
+
+  # From https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route53_query_log
+  # There are restrictions on the configuration of query logging.
+  # Notably, the CloudWatch log group must be in the us-east-1 region,
+  # a permissive CloudWatch log resource policy must be in place,
+  # and the Route53 hosted zone must be public.  <== NOTE
+  # See Configuring Logging for DNS Queries for additional details.
+  # https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/query-logs.html?console_help=true#query-logs-configuring
+
+  # if it has a VPC associated, it's a private DNS zone, so this rule cannot apply because it would require
+  # configuring an invalid logging resources (given the above)
+  not vpc
 }
 
 {{.prefix}}route53LoggingDisabled[route.id] {
--- a/pkg/policies/opa/rego/aws/aws_route53_query_log/route53LoggingDisabled.rego
+++ b/pkg/policies/opa/rego/aws/aws_route53_query_log/route53LoggingDisabled.rego
@@ -2,9 +2,6 @@ package accurics
 
 {{.prefix}}route53LoggingDisabled[route.id] {
   route := input.aws_route53_zone[_]
-  vpc := route.config.dynamic[_].vpc
-  not input.aws_route53_query_log
-
   # From https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route53_query_log
   # There are restrictions on the configuration of query logging.
   # Notably, the CloudWatch log group must be in the us-east-1 region,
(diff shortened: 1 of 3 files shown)


amirbenv:
Thanks for the PR! The commit include changes to go.sum and go.mod, was that intentional? if not, perhaps rebase?

matt-slalom:
> Thanks for the PR! The commit include changes to go.sum and go.mod, was that intentional? if not, perhaps rebase?

Ooops. My bad. Should be fixed now, @amirbenv

pkg/policies/opa/rego/aws/aws_route53_query_log/route53LoggingDisabled.rego
```diff
@@ -2,11 +2,24 @@ package accurics
 
 {{.prefix}}route53LoggingDisabled[route.id] {
   route := input.aws_route53_zone[_]
+  vpc := route.config.dynamic[_].vpc
```
gaurav-gogia, line 5: Hi, 

It looks like object `dynamic` is not a part of normalise.json. I've tested it via terrascan. I referred [terraform docs for route53](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route53_zone) to create the resource. 

Following edit should work: 
`vpc  :=  route.config.vpc`

gaurav-gogia (changes requested):

pkg/policies/opa/rego/aws/aws_route53_query_log/route53LoggingDisabled.rego
matt-slalom: Hi @gaurav-gogia. I'm new to Terrascan, so I apologize in advance if this is a dumb question.

When I use `terrascan scan --config-only -o json` the output includes this stanza. I see in it there's a dynamic object.

Is there a different mechanism to obtain the normalize.json you mentioned?

```
  "aws_route53_zone": [
    {
      "id": "aws_route53_zone.private_zone",
      "name": "private_zone",
      "source": "../../modules/route53/create-dns.tf",
      "line": 125,
      "type": "aws_route53_zone",
      "config": {
        "dynamic": [
          {
            "vpc": {
              "content": [
                {
                  "vpc_id": "${vpc.value.vpc_id}",
                  "vpc_region": "${vpc.value.region}"
                }
              ],
              "for_each": "${local.vpcs}"
            }
          }
        ],
        "for_each": "${var.dns_info}",
        "name": "${each.key}",
        "tags": "${merge(${var.global_tags}, {\n    Name = each.key\n  })}"
      },
      "skip_rules": null
    }
  ]
```

matt-slalom: I tested using `vpc := route.config.vpc` against a private zone and changed `not vpc` to `vpc`. I expect it to call out the zone I have configured, and it does not. However, using the code as I've submitted, the test (changing to `vpc`) works as expected.

It seems `vpc := route.config.vpc` doesn't work?

harkirat22: Hey, thank you for raising the PR, really appreciate it.
So, n=only one change needs to be done. As per latest terraform for aws provider, to create a private zone you don't have to create a dynamic block and then configure vpc block inside it. So, could you please change your terraform file to something: `resource "aws_route53_zone" "private" {
  name = "example.com"

  vpc {
    vpc_id = aws_vpc.example.id
  }
}`
Finally, you can change this `vpc := route.config.dynamic[_].vpc` to `vpc := route.config.vpc`

matt-slalom: I think I see the problem here. Yes, you are correct that we are not required to use a dynamic block, but for those of us who are using it, the change you propose does not work. On the other hand, you raise a good point: my proposed change doesn't work for people who are not using a dynamic block.

I will revise this to account for both scenarios.

matt-slalom: @harkirat22, I've made changes to look for either a static of dynamic VPC block.

harkirat22: Thanks a-lot for covering both the scenarios. It is good to go now.

harkirat22 (approved):

gaurav-gogia (approved):

PR #635, opened 2021-03-30
fix windows tests

patilpankaj212:
1. fix windows tests
2. go mod tidy


pkg/iac-providers/terraform/commons/load-dir.go
```diff
@@ -201,7 +202,7 @@ func processLocalSource(req *hclConfigs.ModuleRequest, remoteModPaths map[string
 
 	pathToModule := req.SourceAddr
 	for p := req.Parent; p != nil; p = p.Parent {
-		pathToModule = filepath.Join(p.SourceAddr, pathToModule)
+		pathToModule = path.Join(p.SourceAddr, pathToModule)
```
dev-gaur, line 205: what's the merit in this change?

patilpankaj212: This was a temporary fix for the unit tests, forgot to add a comment.
This should be changed with other fix.

dev-gaur (approved):

PR #640, opened 2021-04-05
adds support to scan sub folders for terraform iac provider

patilpankaj212:
1. adds support for sub folder scanning for terraform iac provider.
2. the default behaviour for terraform iac provider is changed, to disable recursive sub folder scan use `--non-recursive` flag
3. a new attribute `plan_root` is added in the resource config and violations summary output


pkg/cli/scan.go
```diff
@@ -68,5 +68,6 @@ func init() {
 	scanCmd.Flags().StringVar(&scanOptions.severity, "severity", "", "minimum severity level of the policy violations to be reported by terrascan")
 	scanCmd.Flags().StringSliceVarP(&scanOptions.categories, "categories", "", []string{}, "list of categories of violations to be reported by terrascan (example: --categories=\"category1,category2\")")
 	scanCmd.Flags().BoolVarP(&scanOptions.showPassedRules, "show-passed", "", false, "display passed rules, along with violations")
+	scanCmd.Flags().BoolVarP(&scanOptions.nonRecursive, "non-recursive", "", false, "will scan specified directory and modules used in it")
```
kanchwala-yusuf, line 71: Can we the help message something like `recursively scans directories and modules`

pkg/runtime/executor.go
```diff
@@ -290,3 +306,12 @@ func implementsSubFolderScan(iacType string) bool {
 	}
 	return false
 }
+
+// makeNonRecursiveTrue sets the 'NonRecursiveScan' field to true
+// only terraform iac provider should be passed to this function, else it will panic
+func makeNonRecursiveTrue(iacP iacProvider.IacProvider) iacProvider.IacProvider {
```
kanchwala-yusuf, line 312: Do you think, it would be better to have non-recursive as part of the IacProvider interface rather than setting an option using reflection package.

pkg/iac-providers/terraform/commons/load-dir.go
```diff
@@ -49,76 +48,198 @@ type ModuleConfig struct {
...
+type TerraformDirectoryLoader struct {
+	absRootDir       string
+	nonRecursive     bool
+	remoteDownloader downloader.ModuleDownloader
+	parser           *hclConfigs.Parser
+	errIacLoadDirs   *multierror.Error
+}
+
+// NewTerraformDirectoryLoader creates a new terraformDirectoryLoader
+func NewTerraformDirectoryLoader(rootDirectory string, nonRecursive bool) TerraformDirectoryLoader {
```
kanchwala-yusuf, line 61: Good work!

pkg/runtime/executor.go
```diff
@@ -278,11 +282,15 @@ func (e *Executor) findViolations(results *Output) error {
 }
 
 // implementsSubFolderScan checks if given iac type supports sub folder scanning
-func implementsSubFolderScan(iacType string) bool {
+func implementsSubFolderScan(iacType string, nonRecursive bool) bool {
 	// iac providers that support sub folder scanning
 	// this needs be updated when other iac providers implement
 	// sub folder scanning
-	iacWithSubFolderScan := []string{"all", "k8s", "helm"}
+	if nonRecursive && iacType == "terraform" {
```
kanchwala-yusuf, line 289: I guess, there is no other option but to implement an IaC specific check

pkg/runtime/executor.go
```diff
@@ -278,11 +282,15 @@ func (e *Executor) findViolations(results *Output) error {
...
+		return false
+	}
+
+	iacWithSubFolderScan := []string{"all", "k8s", "helm", "terraform"}
```
kanchwala-yusuf, line 293: Not in this PR, but we should have this information at the time of IaC registration and create a list of iac types which implement recursive scanning in the `init()` of package `iac-providers`. And we can fetch it from there, hence eliminating the need to hard code these values here.

kanchwala-yusuf (approved):

PR #644, opened 2021-04-07
fix - k8s resources config data has absolute source paths for resources

patilpankaj212:
1. k8s iac - filename should be relative
2. fix failing e2e tests

### DIFF
--- a/pkg/iac-providers/kubernetes/v1/load-dir.go
+++ b/pkg/iac-providers/kubernetes/v1/load-dir.go
@@ -43,10 +43,34 @@ func (k *K8sV1) LoadIacDir(absRootDir string) (output.AllResourceConfigs, error)
 			}
 
 			for key := range configData {
+				// the source path formed for each resources is absolute, which should be relative
+				resourceConfigs := configData[key]
+				updateSourcePaths(absRootDir, resourceConfigs)
+
 				allResourcesConfig[key] = append(allResourcesConfig[key], configData[key]...)
 			}
 		}
 	}
 
 	return allResourcesConfig, nil
+}
+
+// updateSourcePaths modifies the source path of each resource from absolute to relative path
+func updateSourcePaths(absRootDir string, resourceConfigs []output.ResourceConfig) {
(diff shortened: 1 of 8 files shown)


kanchwala-yusuf (dismissed):

pkg/iac-providers/kubernetes/v1/load-dir.go
```diff
@@ -43,6 +43,24 @@ func (k *K8sV1) LoadIacDir(absRootDir string) (output.AllResourceConfigs, error)
...
+					// update the source path
+					r.Source, err = filepath.Rel(absRootDir, r.Source)
+
+					// though this error should never occur, but if occurs for some reason, assign the old value source
+					if err != nil {
+						r.Source = oldSource
+						zap.S().Debug("error while getting the relative path of file", zap.String("IAC file", file), zap.Error(err))
+						continue
+					}
+				}
```
kanchwala-yusuf, line 62: Just a suggestion, it would be nice if could move this piece of code into a separate function. then we may be able to unit test as well :)

patilpankaj212: done!

pkg/iac-providers/kubernetes/v1/load-dir.go
```diff
@@ -43,10 +43,34 @@ func (k *K8sV1) LoadIacDir(absRootDir string) (output.AllResourceConfigs, error)
...
 				allResourcesConfig[key] = append(allResourcesConfig[key], configData[key]...)
 			}
 		}
 	}
 
 	return allResourcesConfig, nil
 }
+
+// updateSourcePaths modifies the source path of each resource from absolute to relative path
+func updateSourcePaths(absRootDir string, resourceConfigs []output.ResourceConfig) {
```
jlk, line 59: Maybe rename this to `makeSourcePathRelative`?

jlk (approved):
generally LGTM, maybe tweak name on method to make it more clear.

kanchwala-yusuf (approved):

PR #653, opened 2021-04-12
fix panic for floating point variables

### LINKED ISSUE #652: panic while resolving floating point variable
* terrascan version: 1.4.0
* Operating System: all

### Description

when terraform config file has reference to a floating point variable, terrascan panics while resolving its value.

### What I Did

created a terraform config which uses floating point variable, and ran `terrascan scan`

patilpankaj212:
fixes #652

### DIFF
--- a/pkg/iac-providers/terraform/commons/cty-converters.go
+++ b/pkg/iac-providers/terraform/commons/cty-converters.go
@@ -18,7 +18,6 @@ package commons
 
 import (
 	"fmt"
-	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/zclconf/go-cty/cty"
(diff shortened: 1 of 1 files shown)


pkg/iac-providers/terraform/commons/cty-converters.go
```diff
@@ -58,7 +64,7 @@ func ctyToSlice(ctyVal cty.Value) (interface{}, error) {
 	var val []interface{}
 	var allErrs error
 
-	if strings.Contains(ctyVal.Type().FriendlyName(), "list") {
+	if ctyVal.Type().IsListType() {
```
kanchwala-yusuf, line 67: Just curious, does `IsListType()` give a more reliable and consistent output?

kanchwala-yusuf (approved):

kanchwala-yusuf:
@patilpankaj212 , was wondering if we could add a sample terraform config in our unit tests for testing this particular case?

PR #663, opened 2021-04-14
add authentication with API key for the /logs endpoint

### LINKED ISSUE #662: Add API_KEY to the /logs endpoint for the validating admission webhook
Add authentication using API key for the `/logs` endpoint for listing all the admission requests

kanchwala-yusuf:
fixes #662

### DIFF
--- a/pkg/http-server/routes.go
+++ b/pkg/http-server/routes.go
@@ -37,7 +37,7 @@ func (g *APIServer) Routes(configFile string) []*Route {
 		{verb: "POST", path: versionedPath("/{iac}/{iacVersion}/{cloud}/remote/dir/scan"), fn: h.scanRemoteRepo},
 
 		// k8s webhook Routes
-		{verb: "GET", path: "/k8s/webhooks/logs", fn: h.getLogs},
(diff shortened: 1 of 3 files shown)


pkg/http-server/routes.go
```diff
@@ -37,7 +37,7 @@ func (g *APIServer) Routes(configFile string) []*Route {
 		{verb: "POST", path: versionedPath("/{iac}/{iacVersion}/{cloud}/remote/dir/scan"), fn: h.scanRemoteRepo},
 
 		// k8s webhook Routes
-		{verb: "GET", path: "/k8s/webhooks/logs", fn: h.getLogs},
+		{verb: "GET", path: "/k8s/webhooks/{apiKey}/logs", fn: h.getLogs},
```
patilpankaj212, line 40: missed this earlier, but, I think having `{apiKey}` in the url is not a good idea, as it would be exposed in the url!

should this auth filter be also applied to `/k8s/webhooks/logs/{uid}` route as well?

patilpankaj212: can be taken up later

patilpankaj212 (approved):

PR #665, opened 2021-04-14
Makes saving of admission requests configurable via a config file option

### LINKED ISSUE #664: Make saving of "admission request" configurable via an option in the config file for the validating admission webhook
Currently terrascan saves all the incoming admission requests into it's database. We need to give users an option to choose whether to save admission requests or not. This can be done by a newly introduced flag `save-requests` in the `k8s-admission-control` section of the config file

kanchwala-yusuf:
fixes #664

### DIFF
--- a/pkg/k8s/admission-webhook/validating-webhook.go
+++ b/pkg/k8s/admission-webhook/validating-webhook.go
@@ -241,6 +241,7 @@ func (w ValidatingWebhook) logWebhook(output runtime.Output,
 	var (
 		currentTime             = time.Now()
 		deniedViolationsEncoded string
+		requestBody             string
 	)
 
 	// encode denied violations into a string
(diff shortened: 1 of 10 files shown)


pkg/k8s/admission-webhook/validating-webhook.go
```diff
@@ -310,7 +315,7 @@ type webhookDenyRuleMatcher struct {
 }
 
 // This class should check if one of the violations found is relevant for the specified K8s deny rules
-func (g *webhookDenyRuleMatcher) match(violation results.Violation, denyRules config.K8sDenyRules) bool {
+func (g *webhookDenyRuleMatcher) match(violation results.Violation, denyRules config.K8sAdmissionControl) bool {
```
patilpankaj212, line 318: should we rename `denyRules` to `admissionControl`?

patilpankaj212 (approved):
LGTM!!
But we should add tests for the new field added.

PR #674, opened 2021-04-16
adds support to scan directory with all iac providers in cli mode

patilpankaj212:
1. all iac scan will be used when an iac type is not specified using the -i flag
2. the scan summary will include an additional element `scan_errors`, which will include all the errors that occurred while the iac providers did the folder scan.
4. `scan_errors` is not supported with default human readable output, also it is not part of `--config-only`


pkg/utils/resource.go
```diff
@@ -61,3 +62,25 @@ func GetResourceCount(resourceMapping map[string][]output.ResourceConfig) (count
 	}
 	return
 }
+
+// UpdateResourceConfigs adds a resource of given type if it is not present in allResources
+func UpdateResourceConfigs(resourceType string, resources []output.ResourceConfig, allResources output.AllResourceConfigs) {
```
kanchwala-yusuf, line 67: Hey @patilpankaj212 , 
Does it make sense to have this function and other functions in this file to be moved to the output package itself. Come to think of it these are not `utils` in real sense but specific use cases related to the `AllResourcesConfig` in the output package.

kanchwala-yusuf (approved):

kanchwala-yusuf (approved):

PR #680, opened 2021-04-20
Handle multiple vpc flow logs

matt-slalom:
The current module does not handle cases where multiple VPCs and flow logs are created. This change covers that case, but with a caveat: it only works when running terrascan against tfplan output.

```
terraform plan -out tfplan.out
terraform show -json tfplan.out > tfplan.json
terrascan scan -i tfplan -t aws -f tfplan.json
```

I can supply sample output of what multiple VPCs and flow logs look like.


harkirat22:
@matt-slalom. Yes, can you please provide terraform sample on which you tested this policy. Just incase, I don't miss that exact test case.

matt-slalom:
I've set up a Rego Playground with this info. Found here: https://play.openpolicyagent.org/p/ybeC3Zp5cn

pkg/policies/opa/rego/aws/aws_vpc/vpcFlowLogsNotEnabled.rego
```diff
@@ -28,4 +28,16 @@ flowLogExist(vpc, vpc_input) = exists {
...
-}
\ No newline at end of file
+} else = exists {
+  	x:=vpc[i]
+    flow_log_vpcs_set := { vpc_id | vpc_mapping[x.id]; vpc_id := x.id}
+    flow_log_vpcs_set[x.id]
+    exists = true
+}
+
+vpc_mapping[vpc_id] = flow_vpc_id {
```
harkirat22, line 38: this will break the design pattern, our policy engine can handle one return value from the rule. In your case, even if you look at the rego playground output, it is returning two outputs as if there are two rules in one single policy:

{
    "vpcFlowLogsNotEnabled": [],
    "vpc_mapping": {
        "vpc-0755f0db2302eeece": "vpc-0755f0db2302eeece",
        "vpc-0b346736ba3cfd21d": "vpc-0b346736ba3cfd21d"
    }
}
Here, only vpcFlowLogsNotEnabled is expected.

harkirat22: Moreover, I realised else conditions are not making the policy readable. Along with this I tried to solve the issue addressed in this PR and solves for both cases: tfplan and normal output.
https://play.openpolicyagent.org/p/Zbs4CBUYRk.

matt-slalom:
Hi @harkirat22. I've updated the code as per your suggestions.

matt-slalom:
Here's an updated Playground as well: https://play.openpolicyagent.org/p/Blzt37nnt0

harkirat22:
@matt-slalom, thanks for updating the code. I have tested it, works perfect.

harkirat22 (approved):

PR #683, opened 2021-04-22
adds support to configure dashboard mode in k8s validating webhook

patilpankaj212:
A new attribute `dashboard` is introduced for `k8s-admission-control` in the terrascan config file.
When `dashboard` is true, violations and admission review response would be logged to db, otherwise, the violations would be logged to the console.

### DIFF
--- a/pkg/http-server/webhook-scan_test.go
+++ b/pkg/http-server/webhook-scan_test.go
@@ -17,7 +17,8 @@ import (
 )
 
 func TestUWebhooks(t *testing.T) {
-	testFilePath := filepath.Join("k8s_testdata", "testconfig.json")
+	k8sTestData := "k8s_testdata"
+	testFilePath := filepath.Join(k8sTestData, "testconfig.json")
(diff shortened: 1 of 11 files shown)


pkg/http-server/webhook-scan_test.go
```diff
@@ -114,30 +115,30 @@ func TestUWebhooks(t *testing.T) {
...
-			configFile:         "./k8s_testdata/config-specific-rule.toml",
+			configFile:         filepath.Join(k8sTestData, "config-specific-rule.toml"),
 			warnings:           false,
 			allowed:            true,
 			wantStatus:         http.StatusOK,
 		},
 		{
 			name:               "risky request object with config that just removes some of the violations",
-			contentRequestPath: "./k8s_testdata/risky_testconfig.json",
+			contentRequestPath: filepath.Join(k8sTestData, "risky_testconfig.json"),
```
kanchwala-yusuf, line 128: Thank you for improving the tests!

kanchwala-yusuf (approved):

Issue #711, opened 2021-04-29
accurics.azure.AKS.3 is defective


xortim:
* terrascan version: 1.50
* Operating System: all

### Description

Using terrascan to interrogate a Terraform plan file. When attempting to resolve the HIGH severity error produced by terrascan with regard to resource locks on azure resources, the provided rule does not compare the relevant components to determine if a lock is present.

### What I Did

`main.tf`
```hcl
locals {
  name_prefix = "prefix"
  name_suffix = "suffix"
}

resource "azurerm_container_registry" "this" {
  name                = "${local.name_prefix}acr${local.name_suffix}"
  resource_group_name = "resource_group_name"
  location            = "eastus"

  sku                      = var.sku
  admin_enabled            = var.admin_enabled
}

resource "azurerm_management_lock" "this" {
  name       = "${local.name_prefix}acr${local.name_suffix}"
  scope      = azurerm_container_registry.this.id
  lock_level = "CanNotDelete"
  notes      = "Managed by terraform. Locked for the warm and fuzzies"
}
```

This rule is trigged as a violation [containerRegistryResourceLock.rego](https://github.com/accurics/terrascan/blob/master/pkg/policies/opa/rego/azure/azurerm_container_registry/containerRegistryResourceLock.rego)

I enabled debugging mode and grabbed the "expected" result and decoded it.

```json
{
    "reme_containerRegistryResourceLock": [
        {
            "Actual": null,
            "Attribute": "",
            "AttributeDataType": "resource",
            "CodeType": "resource",
[...]

PR #712, opened 2021-04-29
Fix accurics.azure.AKS.3

xortim:
Resolves https://github.com/accurics/terrascan/issues/711


amirbenv:
Thanks for the fix! will review shortly

harkirat22 (approved):
LGTM

PR #713, opened 2021-04-29
update version

chenrui333:
relates to Homebrew/homebrew-core#76294

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -17,7 +17,7 @@
 package version
 
 // Terrascan The Terrascan version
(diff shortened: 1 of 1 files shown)


jlk (approved):
LGTM, thx!

jlk:
Thanks @chenrui333. We're working on a release checklist so we don't make basic mistakes like this.

chenrui333:
> Thanks @chenrui333. We're working on a release checklist so we don't make basic mistakes like this.

no worries at all. :)

Issue #718, opened 2021-04-30
terrascan "latest" docker image broken for tfplan


xortim:
* terrascan version: 1.5.1
* Operating System: linux

### Description

Running the same command fails between tagged 1.5.1 and 'latest'

### What I Did

```sh
docker run --rm -v $(pwd):/workspace --workdir /workspace accurics/terrascan:1.5.1 scan -i tfplan -f /workspace/tfplan.out.json
...
Scan Summary -

	File/Folder         :	/workspace/tfplan.out.json
	IaC Type            :	tfplan
	Scanned At          :	2021-04-30 18:41:24.542798099 +0000 UTC
	Policies Validated  :	562
	Violated Policies   :	2
	Low                 :	1
	Medium              :	0
	High                :	1
```

```sh
docker run --rm -v $(pwd):/workspace --workdir /workspace accurics/terrascan:latest scan -i tfplan -f /workspace/tfplan.out.json
2021-04-30T18:42:43.959Z	error	cli/run.go:110	scan run failed{error 26 0  1 error occurred:
	* tfplan should always be a file, not a directory. Please specify path to tfplan file  with '-f' option
```

jlk:
@xortim we've identified a fix for that, will have it merged in later today hopefully.

jlk:
@xortim - didn't mean to autoclose this with the PR merge. Let us know if this is still an issue.

PR #720, opened 2021-04-30
fix - #718

patilpankaj212:
- The value for iac directory remains `.`, when iac file path is specified.
- Added a unit test for this scenario in run_test.go


jlk (approved):
LGTM!

PR #724, opened 2021-05-03
changes for argocd integration

patilpankaj212:
1. modify Dockerfile to add a folder .ssh in /home/terrascan. This folder will be used to place ssh keys, ssh config and known_hosts required for api server to download repositories.
2. modify the resources skipping for kubernetes
3. modify response code of api server's remote repo scan response, when the result contains admission controller denied violations.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -27,6 +27,10 @@ RUN addgroup --gid 101 terrascan && \
     adduser -S --uid 101 --ingroup terrascan terrascan && \
     apk add --no-cache git openssh
 
+# create .ssh folder and change owner to terrascan
+RUN mkdir -p /home/terrascan/.ssh && \
+    chown -R terrascan:terrascan /home/terrascan/.ssh
+
 # run as non root user
 USER terrascan
 
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -27,9 +27,9 @@ RUN addgroup --gid 101 terrascan && \
     adduser -S --uid 101 --ingroup terrascan terrascan && \
     apk add --no-cache git openssh
 
-# create .ssh folder and change owner to terrascan
-RUN mkdir -p /home/terrascan/.ssh && \
-    chown -R terrascan:terrascan /home/terrascan/.ssh
+# create ~/.ssh & ~/bin folder and change owner to terrascan
+RUN mkdir -p /home/terrascan/.ssh /home/terrascan/bin && \
+    chown -R terrascan:terrascan /home/terrascan
 
 # run as non root user
 USER terrascan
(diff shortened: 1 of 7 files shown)


pkg/iac-providers/kubernetes/v1/normalize.go
```diff
@@ -159,47 +159,16 @@ func readSkipRulesFromAnnotations(annotations map[string]interface{}, resourceID
...
-	// get rule, if rule not found return nil
-	if rule, ok = m[terrascanSkipRule]; ok {
-		if _, ok = rule.(string); ok {
-			skipRule.Rule = rule.(string)
-		} else {
+	if rules, ok := skipRulesFromAnnotations.(string); ok {
+		skipRules := make([]output.SkipRule, 0)
+		err := json.Unmarshal([]byte(rules), &skipRules)
+		if err != nil {
+			zap.S().Debugf("terrascanSkip rules '%s' cannot be unmarshalled to json", rules)
```
dev-gaur, line 166:
```suggestion
			zap.S().Debugf("json string %s cannot be unmarshalled to [ ]output.SkipRules struct schema", rules)
```

dev-gaur (changes requested):
One very minor change requested.

build/Dockerfile
```diff
@@ -27,6 +27,10 @@ RUN addgroup --gid 101 terrascan && \
     adduser -S --uid 101 --ingroup terrascan terrascan && \
     apk add --no-cache git openssh
 
+# create .ssh folder and change owner to terrascan
+RUN mkdir -p /home/terrascan/.ssh && \
+    chown -R terrascan:terrascan /home/terrascan/.ssh
+
```
jlk, line 33: That needs to be mode 700 as well. umask might already be set, but keep in mind...

pkg/iac-providers/kubernetes/v1/normalize.go
jlk: Devang's comment is technically accurate, but not sure if that much detail would be useful to average user, or confusing?

jlk (dismissed):
Generally LGTM. Note the 2 comments, but not requiring any changes...

pkg/iac-providers/kubernetes/v1/normalize.go
amirbenv: I agree with @jlk. This change will not be informative to most users.

jlk (approved):
Ship it! :)

pkg/iac-providers/kubernetes/v1/normalize.go
dev-gaur: Debug is also used by us developers.. to diagnose what went wrong. Will ease it for us to help people reporting bugs and other issues.

dev-gaur (approved):

PR #727, opened 2021-05-03
Fix NSG associations

### LINKED ISSUE #725: accurics.azure.NS.161 does not work with tfplan
* terrascan version: 1.5.0
* Operating System: macOS, Linux

### Description

Attempting to use Terrascan against a Terraform Plan file containing Azure VNets, subnets, NSG, and the NSG association resource. Terrascan does not take into account that plan contains the actual resource ID and not the internal HCL name. Similar problem to #711. It instead, tells me that every subnet does not contain an association. There's no Rego present that would inspect the association using the actual resource IDs.

xortim:
- Format file with `opa fmt`

Resolves #725


harkirat22 (approved):
LGTM.

Issue #709, opened 2021-04-29
Failed to run prepared query error in opa/engine.go


henrychoi7:
* terrascan version: 1.5.0
* Operating System: macOS 11.2.3

### Description

I was trying to test it on my Terraform code but an error keeps occur every moment when I execute it. I think it should be going without any errors, but every code I test it with makes the same error.

`2021-04-29T16:53:45.436+0900	warn	opa/engine.go:346	failed to run prepared query{error 26 0  hardCodedShellScript:22: eval_builtin_error: base64.decode: illegal base64 data at input byte 0} {rule 15 0 'AWS.LaunchConfiguration.DataSecurity.High.0101' <nil>} {file 15 0 hardCodedShellScript.rego <nil>}`

### What I Did

Example Terraform:

```terraform
provider "aws" {
  region = "eu-west-1"
}

locals {
  user_data = <<EOF
#!/bin/bash
echo "Hello Terraform!"
EOF
}
data "aws_vpc" "default" {
  default = true
}

data "aws_subnet_ids" "all" {
  vpc_id = data.aws_vpc.default.id
}

data "aws_ami" "amazon_linux" {
  most_recent = true

  owners = ["amazon"]

  filter {
    name = "name"

    values = [
      "amzn-ami-hvm-*-x86_64-gp2",
    ]
  }

  filter {
    name = "owner-alias"

    values = [
      "amazon",
    ]
  }
}

module "security_group" {
  source  = "terraform-aws-modules/security-group/aws"
  version = "~> 3.0"

  name        = "example"
  description = "Security group for example usage with EC2 instance"
  vpc_id      = data.aws_vpc.default.id

  ingress_cidr_blocks = ["0.0.0.0/0"]
  ingress_rules       = ["http-80-tcp", "all-icmp"]
  egress_rules        = ["all-all"]
}

resource "aws_eip" "this" {
  vpc      = true
  instance = module.ec2.id[0]
}

resource "aws_placement_group" "web" {
  name     = "hunky-dory-pg"
  strategy = "cluster"
}

resource "aws_kms_key" "this" {
}

module "ec2" {
  source = "../../"

  instance_count = 1

  name          = "example-normal"
  ami           = data.aws_ami.amazon_linux.id
  instance_type = "c5.large"
  subnet_id     = tolist(data.aws_subnet_ids.all.ids)[0]
  vpc_security_group_ids      = [module.security_group.this_security_group_id]
  associate_public_ip_address = true
[...]

PR #770, opened 2021-05-11
add check for unresolved variables

gaurav-gogia:
- adds a check for `$` in the beginning of `user_data_base64` to account for unresolved variables
- works as a workaround for base64 decode error caused due to incorrect string
- fixes issue #709


kanchwala-yusuf:
This issue needs to be fixed from terrascan side, closing this PR. We can reopen if required!

Issue #766, opened 2021-05-11
Update Docker build for terrascan to use numeric UID


jlk:
* terrascan version: 1.5.0
* Operating System: MacOS

### Description

When developing helm chart, to follow best practices I'd like to set a securityContext of `runAsNonRoot: true`
Unfortunately, when the k8s deployment creates a pod, the following error is seen in pod details:

```
Events:
  Type     Reason     Age               From               Message
  ----     ------     ----              ----               -------
  Normal   Scheduled  13s               default-scheduler  Successfully assigned default/terrahook-7b8fd679ff-jsk5k to aks-agentpool-30212186-vmss000002
  Normal   Pulling    13s               kubelet            Pulling image "alpine/git"
  Normal   Pulled     12s               kubelet            Successfully pulled image "alpine/git"
  Normal   Created    12s               kubelet            Created container git-cloner
  Normal   Started    12s               kubelet            Started container git-cloner
  Normal   Pulled     9s (x2 over 10s)  kubelet            Container image "accurics/terrascan:1.5.0" already present on machine
  Warning  Failed     9s (x2 over 10s)  kubelet            Error: container has runAsNonRoot and image has non-numeric user (terrascan), cannot verify user is non-root
```

So looks like we need to change the Dockerfile to read
```
USER 101
```

PR #773, opened 2021-05-12
Modified the Dockerfile to use numeric UID 

Rchanger:
#766 
Fix: modified the docker file to use numeric UID.
```
USE 101
``` 
Reason for change:
 
when we set the RunAsNonRoot field of a security context, k8s internal logic checks the container for numeric UID instead of a username, and if it doesnot found any numeric UID it throughs error `container has runAsNonRoot and image has non-numeric user (terrascan), cannot verify user is non-root
`

Tested the changes for Use cases

- Admission controller webhook with plane YAML files
-  Admission controller with Helm
- Admission controller with argoCD

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -32,7 +32,7 @@ RUN mkdir -p /home/terrascan/.ssh /home/terrascan/bin && \
     chown -R terrascan:terrascan /home/terrascan
 
 # run as non root user
-USER terrascan
+USER 101
 
 ENV PATH /go/bin:$PATH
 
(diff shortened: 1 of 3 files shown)


dev-gaur:
Built the docker file and by default `whoami` returns 'terrascan'.

```
$ docker exec -it c7dfa4829a14 /bin/sh
/ $ whoami
terrascan
/ $ ps aux
PID   USER     TIME  COMMAND
    1 terrasca  0:03 /go/bin/terrascan server --log-type json
   17 terrasca  0:00 /bin/sh
   24 terrasca  0:00 ps aux
```

dev-gaur (approved):

Issue #735, opened 2021-05-05
Issue in GCP Policyfile unrestrictedRdpAccess.rego


menzbua:
* terrascan version: 1.5.1
* Operating System: Linux

I think i found a bug in the File terrascan/pkg/policies/opa/rego/gcp/google_compute_firewall/unrestrictedRdpAccess.rego. In my environment i got the violation that i have unrestricted RDP Access Allowed to my infrastructure but we have restricted the source addresses in the rules. So for my perspective there should be an addition to the policy file

```fire_rule.source_ranges[_] == "0.0.0.0/0"```

After i added that to in my environment, i really get the violations for rules that have unrestricted access to my servers via RDP.

What do you think about it?

Regards
Manuel

**Comment by gaurav-gogia** [comment]
hi @menzbua, can you please share terraform file for the same? it'll be easier to understand the bug that way.

**Comment by menzbua** [comment]
hi @gaurav-gogia,

for example this is a terraform code that terrascan detects as unrestricted rdp access:

```
resource "google_compute_firewall" "" {
  name               = "ingress-allow-rdp"
  description        = "terraform managed"
  disabled           = false
  direction          = "INGRESS"
  network            = google_compute_network
  priority           = 1000
  source_ranges      = ["27.245.20.0/20"]

  allow {
    protocol = "tcp"
    ports    = ["3389"]
  }
}
```

Regards
Manuel

gaurav-gogia:
@menzbua 

Thanks! I understand the bug in policy code now. 

Would you like to raise a PR with the suggested fix or should we move forward with it?

menzbua:
@gaurav-gogia 

I will create a Pull Request with the fix.

Regards
Manuel

PR #776, opened 2021-05-12
Added source_range 0.0.0.0/0 (any) to avoid rule violations

menzbua:
Added source_range 0.0.0.0/0 to unrestrictedRdpAccess.rego because you will get rule violations when you have restricted RDP Access zu your environment with source_range set.

See issue #735


gaurav-gogia (approved):
Looks good to me. @harkirat22 could you also please review when you get time?

PR #779, opened 2021-05-12
Automate generation of TLS Certs using Helm

rahulchheda:
In this PR, I tried to create self-signed certificate for Terrascan Admission Controller using Helm. This seems to be a very simple implementation for this.

Signed-off-by: Rahul M Chheda <rahul.chheda@accurics.com>


dev-gaur:
@rahulchheda can you update the README there as well?

deploy/helm-charts/templates/secret.yaml
```diff
@@ -1,11 +1,22 @@
...
 metadata:
   name: {{ .Values.secret_name }}
   namespace: {{ .Release.Namespace }}
-type: Opaque
 data:
-  key: |-
-    {{ .Files.Get "data/server.key" | b64enc }}
-  cert: |-
-    {{ .Files.Get "data/server.crt" | b64enc }}
+{{- if and (eq "" .Values.secrets.tlsCertFilePath) (eq "" .Values.secrets.tlsKeyFilePath) }}
```
dev-gaur, line 8: either we should use `or` operator here or we should have additional else if cases that handle the cases where key is blank but cert is defined and vice versa.

deploy/helm-charts/values.yaml
```diff
@@ -1,7 +1,13 @@
...
 terrascan_container_image: accurics/terrascan:1.6.0
 use_debug: true
-secret_name: terrascancerts
+
+# provide secrets for admission controller
+secrets:
+  tlsCertFilePath: ""
+  tlsKeyFilePath: ""
+
+secret_name: "terrascancerts"
```
dev-gaur, line 10: do we need the quotes here?

dev-gaur (changes requested):
Thanks for your contribution.

minor changes requested. please update the README.md for this feature.

deploy/helm-charts/templates/secret.yaml
rahulchheda: `and` would do the trick in this case, if any of the 2 keys are missing, it will error out somewhat like this: `unknown object type "nil" in Secret.data.tls.crt`

dev-gaur: great!

dev-gaur (approved):
LGTM. Working for me.

PR #780, opened 2021-05-13
config file changes for terrascan server

patilpankaj212:
1. fix issue with global config
2. make changes in server code to remove reading config file
3. unit tests

### DIFF
--- a/pkg/config/global.go
+++ b/pkg/config/global.go
@@ -41,6 +41,8 @@ var (
 // in configFile will get default values
 func LoadGlobalConfig(configFile string) error {
 	// Start with the defaults
+	global = &TerrascanConfig{}
+
 	global.Policy = Policy{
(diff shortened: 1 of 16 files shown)


dev-gaur:
Is it a refactor? or a change in config? I don't see any changes to the TerrascanConfig struct. Its probably a refactor as per mey understanding.

@patilpankaj212

pkg/config/global.go
```diff
@@ -120,50 +122,80 @@ func LoadGlobalConfig(configFile string) error {
 
 // GetPolicyBasePath returns the configured policy base path
 func GetPolicyBasePath() string {
+	if global == nil {
```
kanchwala-yusuf, line 125: Can there be a case where `global` is not nil? But, `global.Policy.BasePath`is an empty string?

kanchwala-yusuf (approved):

Issue #782, opened 2021-05-13
Scan for terraform doesn't error out if a module definition refers to a directory with no tf files


patilpankaj212:
* terrascan version: 1.6.0
* Operating System: All

### Description
If a module definition in a terraform file has `source` as a local directory which doesn't contain any tf config files, scan should result in an error.

### What I Did

```
terrascan scan -i terraform -d <dir> --non-recursive
```

The directory contains a tf file that defines a module that has a local source.

```
module "ec2" {
  source = "../../"

  instance_count = 1

  ...
}

```

PR #793, opened 2021-05-19
fix: added validation for module local source dir

Rchanger:
#782 
Fix: Added check for the module source local directory.


pkg/iac-providers/terraform/commons/load-dir.go
```diff
@@ -321,6 +321,7 @@ func (t TerraformDirectoryLoader) buildUnifiedConfig(rootMod *hclConfigs.Module,
 			// figure out path sub module directory, if it's remote then download it locally
 			var pathToModule string
 			var err error
+			var moduleDirDiags hcl.Diagnostics
```
kanchwala-yusuf, line 324: These can go in a single var block, just makes the code tidier

pkg/iac-providers/terraform/commons/load-dir.go
```diff
@@ -321,6 +321,7 @@ func (t TerraformDirectoryLoader) buildUnifiedConfig(rootMod *hclConfigs.Module,
...
 			if downloader.IsLocalSourceAddr(req.SourceAddr) {
```
kanchwala-yusuf, line 325: Wondering, if it is time to get rid of this if-else ladder and move to a switch case?

kanchwala-yusuf (approved):
Have some minor comments, approving this PR

patilpankaj212 (approved):

pkg/iac-providers/terraform/commons/load-dir.go
Rchanger: I'll make this change .

pkg/iac-providers/terraform/commons/load-dir.go
```diff
@@ -344,6 +347,15 @@ func (t TerraformDirectoryLoader) buildUnifiedConfig(rootMod *hclConfigs.Module,
 				}
 			}
 
+			// verify whether the module source directory has any .tf config files
+			if !t.parser.IsConfigDir(pathToModule) {
```
patilpankaj212, line 351: we should modify this check because this will also modify the error message for non existing directories.

patilpankaj212 (changes requested):

patilpankaj212 (approved):

kanchwala-yusuf (approved):

Issue #805, opened 2021-05-24
Add ID field


gaurav-gogia:
A new field needs to be added in rule metadata to indicate the use of latest ID format. 

This new format is:
AC_PolicType_FourDigitNumber

PR #808, opened 2021-05-24
Feature/az id field

gaurav-gogia:
Changes for terrascan policies where names match with SIAC policies

These changes are relevant only for Azure policies

1. Add `id` field with new format
2. Update severity
3. Update category

PR created for issue #805


gaurav-gogia:
I see that there's a big disparity between network security rules of azure and siac, that may have caused some issues like id collision. I'll try find resolution for the same.

nasir-rabbani (approved):

PR #813, opened 2021-05-26
Remove duplicate securityContext field from job template definition

dev-gaur:


integrations/argocd/examples/presynchook-webhook.yml
```diff
@@ -4,14 +4,11 @@ metadata:
  generateName: terrascan-hook-
  namespace: <YOU APPLICATION NAMESPACE>
  annotations:
-   argocd.argoproj.io/hook: PreSync            
+   argocd.argoproj.io/hook: PreSync
 spec:
  ttlSecondsAfterFinished: 3600
  template:
    spec:
-     securityContext:
```
Rchanger, line 12: Actually, that is not a duplicate field we can have a security context at Pod level as well as container level, but since we have only one container inside the pod we can remove that Pod level security context. But I have observed that if we removed this field terrascan gives violation.
```
 {
        "rule_name": "securityContextUsed",
        "description": "Apply Security Context to Your Pods and Containers",
        "rule_id": "AC-K8-NS-PO-M-0122",
        "severity": "MEDIUM",
        "category": "Infrastructure Security",
        "resource_name": "terrascan-hook-",
        "resource_type": "kubernetes_job",
        "file": "/Users/suvarna/go/src/github.com/rchanger/terrascan/integrations/argocd/examples/presynchook.yml",
        "line": 1
      },
```

So for now we can keep this field and after policy fixes, we can remove it. It does not have any impact as well

Rchanger:
Actually, that is not a duplicate field we can have a security context at Pod level as well as container level, but since we have only one container inside the pod we can remove that Pod level security context. But I have observed that if we removed this field terrascan gives violation.
```
 {
        "rule_name": "securityContextUsed",
        "description": "Apply Security Context to Your Pods and Containers",
        "rule_id": "AC-K8-NS-PO-M-0122",
        "severity": "MEDIUM",
        "category": "Infrastructure Security",
        "resource_name": "terrascan-hook-",
        "resource_type": "kubernetes_job",
        "file": "/Users/suvarna/go/src/github.com/rchanger/terrascan/integrations/argocd/examples/presynchook.yml",
        "line": 1
      }
```
So for now we can keep this field and after policy fixes, we can remove it. It does not have any impact as well.
Closing the PR since the changes are not required.

PR #817, opened 2021-05-26
[feat.] Merge Webhook and Server Helm Chart

rahulchheda:
Tried to merge Webhook and Server charts in this PR. We really don't need to maintain them separately.
PTAL, and let me know if any improvements are needed.

Signed-off-by: Rahul M Chheda <rahul.chheda@accurics.com>


deploy/helm-charts/server/values.yaml
```diff
@@ -8,3 +8,22 @@ name: terrascan
 persistence:
   enabled: false
   existingclaim: terrascanPvc
+
+webhook:
+  name: webhook.terrascan.io
+  failurePolicy: Ignore
```
rahulchheda, line 14: Had to change this to `Ignore`, because if service is un-available for webhook, it should block all the specified resources in cluster.
Hope this is the desired behaviour.
cc: @dev-gaur @kanchwala-yusuf @jlk

deploy/helm-charts/server/templates/service.yaml
```diff
@@ -4,7 +4,6 @@ metadata:
   name: {{ .Values.name }}
   namespace: {{ .Release.Namespace }}
 spec:
-  type: LoadBalancer
```
rahulchheda, line 7: No need to create `LoadBalancer` service, until we are using this service from outside of this cluster. It's the user choice to change it afterwards to enhance its usecases.

deploy/helm-charts/server/templates/policy_fail_vw.yaml
```diff
@@ -0,0 +1,63 @@
+{{- if eq .Values.webhook.failurePolicy "Fail" }}
```
rahulchheda, line 1: Had to create this file just to support `validatingwebhookconfiguration` failurePolicy to be `FAIL`.

It turns out, webhook doesn't allow the terrascan server pod to come up in case `failurePolicy` is `Fail`.

So, as a workaround, we create the webhook w/ `Ignore`, and then upgrade it to `Fail` in. post install chart hook. ref: https://v2.helm.sh/docs/charts_hooks/

dev-gaur:
@rahulchheda sent a PR to your branch for all the remaining changes.

https://github.com/rahulchheda/terrascan/pull/1

deploy/helm-charts/server/templates/policy_fail_vw.yaml
dev-gaur: > Had to create this file just to support validatingwebhookconfiguration failurePolicy to be FAIL.

It turns out, webhook doesn't allow the terrascan server pod to come up in case failurePolicy is Fail.

So, as a workaround, we create the webhook w/ Ignore, and then upgrade it to Fail in. post install chart hook. ref: https://v2.helm.sh/docs/charts_hooks/

Can you add this as a comment in the vw yaml file as well?

rahulchheda: sure

dev-gaur (approved):

PR #825, opened 2021-05-31
Do not trim resource id from tfplan json

### LINKED ISSUE #702: tfplan should use resource address for id field
* terrascan version: 1.5.0
* Operating System: macOS

### Description

I'm attempting to identify which resources are in violation of policies. Being able to identify resources in the same or similar way as Terraform's own output and internal representation of these resources would be helpful. So instead of `null_resource.this` we would get `module.example.null_resource.this`. It would be helpful to have a better understanding of how the different resource representations within the plan file are merged by terrascan.

### What I Did

`./main.tf`
```hcl
resource "null_resource" "this" {
  provisioner "local-exec" {
    command = "echo 'Hello'"
  }
}
```

`./example/main.tf`
```hcl
module "example" {
  source "../"
}
```

`terrascan scan -i tfplan --config-only -o json`
```json
{
  "null_resource": [
    {
      "id": "null_resource.this",
      "name": "this",
      "source": "",
      "line": 0,
      "type": "null_resource",
      "config": {
        "triggers": null
      },
      "skip_rules": null
    }
  ]
}
```

In the plan file, these have the desired attribute under `planned_values` and `resource_changes`:
```json
{
  "format_version": "0.1",
  "terraform_version": "0.14.10",
  "planned_values": {
    "root_module": {
      "child_modules": [
        {
          "resources": [
            {
              "address": "module.example.null_resource.this",
...
  "resource_changes": [
    {
      "address": "module.example.null_resource.this",
...
```

kanchwala-yusuf:
fixes #702 
Resource ID in the tfplan json is of the format: `module.acqa-test-cbmodule1.aws_codebuild_project.cb_project`, but in terrascan we trim it to `aws_codebuild_project.cb_project`.

As a result, it becomes difficult to identify module in which the resource has been created.

This PR fixes this problem


patilpankaj212 (approved):
LGTM!!
The resourceID change for k8 seems harmless as it is internal to terrascan.

PR #833, opened 2021-06-03
Support for spaces in policy reference_id

nasir-rabbani:
1. Updated regex to skip rules with spaces in reference_id.
2. Added test cases for reference_id with spaces

### DIFF
--- a/pkg/utils/skip_rules.go
+++ b/pkg/utils/skip_rules.go
@@ -18,6 +18,7 @@ package utils
 
 import (
 	"encoding/json"
+	"fmt"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/utils/skip_rules.go
```diff
@@ -35,15 +35,17 @@ const (
...
+	skipRulesPattern               = regexp.MustCompile(`(#ts:skip=` + ruleNameRegex + `([ \t]+.*){0,1})`)
 	skipRulesPrefix                = "#ts:skip="
 	infileInstructionNotPresentLog = "%s not present for resource: %s"
 )
 
 // GetSkipRules returns a list of rules to be skipped. The rules to be skipped
 // can be set in terraform resource config with the following pattern:
 // #ts:skip=AWS.S3Bucket.DS.High.1043
-// $ts:skip=AWS.S3Bucket.DS.High.1044 reason to skip the rule
+// #ts:skip=AWS.S3Bucket.DS.High.1044 reason to skip the rule
```
dev-gaur, line 48: does this change has any relevance or significance?

nasir-rabbani: For relevance to the issue, I think no. But for significance, if we read the comment on line 46 it might create confusion for the reader that we accept # and $ both for skipping rules.

dev-gaur (changes requested):
I'd be great if you could also add some e2e tests for this. since this is a functional change/enhancement.

patilpankaj212 (changes requested):
Hi @nasir-rabbani, I think we can now add test in the policies package and verify that all the reference id's / id's pass the regex.

pkg/utils/skip_rules.go
dev-gaur: okay then

dev-gaur (approved):

patilpankaj212 (approved):
we can push this PR for now, but when we add tests for policies, we would need to ensure that spaces are not allowed in the `rule ID` / `reference ID`.

PR #845, opened 2021-06-08
reference ids with & and <space> fixed

gaurav-gogia:
- some reference ids had `&` symbol and whitespace ` `, they have been taken care of


nasir-rabbani (approved):
LGTM

PR #846, opened 2021-06-08
Bugfix/use ref id old format

gaurav-gogia:
- Fixed `reference_id` for those json files that were using incorrect format
- These fixed `reference_id` fields try to use old format as closely as possible


nasir-rabbani (approved):
LGTM

PR #865, opened 2021-06-15
Change api, Add support for s3 bucket resource and better cft loader

sigmabaryon:
Changed iac loader
Added Load-file tests
Improved Load dir tests not right


pkg/iac-providers/cft/v1/load-dir_test.go
```diff
@@ -107,23 +113,48 @@ func TestLoadIacDir(t *testing.T) {
...
+			if err != nil {
+				t.Errorf("error loading output.json ResourceConfig, %T", err)
+			}
+
+			// check if resourcetype and resources are present
+			for name, resources := range testArc {
+				if allResourceConfigs[name] == nil {
+					t.Errorf("resource Type %s from test data %s not found", name, resourceDir)
+				}
+				for _, testResource := range resources {
```
patilpankaj212, line 144: Hey @sigmabaryon, I was wondering if `deepEqual` would have worked here? Because we are validating if the `ID` is present or not, but not validating other fields of the config resource.

pkg/iac-providers/cft/v1/load-dir_test.go
```diff
@@ -107,23 +113,48 @@ func TestLoadIacDir(t *testing.T) {
...
+							found = true
+							break
+						}
+					}
+					if !found {
+						t.Errorf("resource %s of type %s from test data not found", testResource.ID, name)
+					}
+				}
+			}
+
```
patilpankaj212, line 157: We might have added some tests for error scenarios !!

patilpankaj212 (approved):

PR #900, opened 2021-06-29
fixes: recursive loop when parent and child module has same local block

### LINKED ISSUE #851: Recursive Loop Scanning Terraform
* terrascan version: v1.16.0
* Operating System: MacOS/Docker

### Description

Terraform scan of a file that contains a module passing a parameter that uses the same name as the local that is used for the same function inside the module.

This generates a recursive loop within Terrascan.

### What I Did

I ran in debug mode and could see that it was recursively expanding the internal local variable. I isolated the error into the code in the attached zip file.

[locals_loop_error.zip](https://github.com/accurics/terrascan/files/6624667/locals_loop_error.zip)

This is similar to a previous error that was fixed.

Rchanger:
fixes #851 

- resolution of the local variable from the parent module
- resolution of reference of map type.

### DIFF
--- a/pkg/iac-providers/terraform/v14/testdata/tfjson/recursive-loop-duplicate-locals.json
+++ b/pkg/iac-providers/terraform/v14/testdata/tfjson/recursive-loop-duplicate-locals.json
@@ -1,0 +1,20 @@
+{
+    "aws_iam_user": [
+      {
+        "id": "aws_iam_user.lb",
+        "name": "lb",
+        "module_name": "dummy",
+        "source": "dummy/main.tf",
+        "plan_root": "./",
+        "line": 15,
+        "type": "aws_iam_user",
+        "config": {
+          "name": "joe != \"\" ? \"joe-fred\" : fred",
+          "tags": "${merge(${merge({\n          CreatedBy = \"Terraform\"\n        }, {\"this\":\"that\"})},\n  {\n    Name = joe != \"\" ? \"joe-fred\" : fred\n  })}"
+        },
+        "skip_rules": null,
+        "max_severity": "",
+        "min_severity": ""
+      }
+    ]
+  }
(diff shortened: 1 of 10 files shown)


pkg/iac-providers/terraform/v14/testdata/tfjson/recursive-loop-duplicate-locals.json
```diff
@@ -0,0 +1,20 @@
...
+        "id": "aws_iam_user.lb",
+        "name": "lb",
+        "module_name": "dummy",
+        "source": "dummy/main.tf",
+        "plan_root": "./",
+        "line": 15,
+        "type": "aws_iam_user",
+        "config": {
+          "name": "joe != \"\" ? \"joe-fred\" : fred",
+          "tags": "${merge(${merge({\n          CreatedBy = \"Terraform\"\n        }, {\"this\":\"that\"})},\n  {\n    Name = joe != \"\" ? \"joe-fred\" : fred\n  })}"
```
dev-gaur, line 13: can you change it to how you changed this part https://github.com/accurics/terrascan/pull/900/files#diff-9e7e2f86151ba69659926d94f991b79ee8c8e03d62733d1aa2dd72207485282aR406 ?

dev-gaur (changes requested):

pkg/iac-providers/terraform/v14/testdata/tfjson/recursive-loop-duplicate-locals.json
Rchanger: That is not possible because here it is a string with an expression that needs to be evaluated. Once we add support for functions this will also change.

kanchwala-yusuf (approved):

dev-gaur (approved):

patilpankaj212 (approved):

PR #907, opened 2021-07-01
add github-sarif writer for github suited sarif output

dev-gaur:


pkg/writer/sarif.go
```diff
@@ -65,19 +69,24 @@ func SarifWriter(data interface{}, writer io.Writer) error {
...
-			return err
+		var artifactLocation *sarif.ArtifactLocation
+
+		if forGithub {
+			artifactLocation = sarif.NewSimpleArtifactLocation(violation.File).
+				WithUriBaseId(outputData.Summary.ResourcePath)
+		} else {
+			absFilePath, err := getAbsoluteFilePath(outputData.Summary.ResourcePath, violation.File)
+			if err != nil {
+				return errors.Errorf("unable to create absolute path, error: %v", err)
```
kanchwala-yusuf, line 83: Is it possible to use `fmt.Errorf()` here?

kanchwala-yusuf (approved):

pkg/writer/sarif.go
dev-gaur: my bad. will correct this in future

PR #630, opened 2021-03-24
fix kustomize scan breakage on windows

dev-gaur:

PR #932, opened 2021-07-16
Fixes supported Kustomize version (should be v4)

dev-gaur:
Since #630 was merged, we had been using v4 of kustomize but declaring it as v3 support . This PR aims to fix that.


pkg/iac-providers/kustomize/v4/load-dir.go
```diff
@@ -23,8 +23,8 @@ var (
 )
 
 // LoadIacDir loads the kustomize directory and returns the ResourceConfig mapping which is evaluated by the policy engine
-func (k *KustomizeV3) LoadIacDir(absRootDir string, options map[string]interface{}) (output.AllResourceConfigs, error) {
```
dev-gaur, line 26:
```suggestion
// LoadIacDir loads the kustomize directory and returns the ResourceConfig mapping which is evaluated by the policy engine
```

nasir-rabbani (approved):

PR #949, opened 2021-07-27
Add support for CFT nested stacks

sigmabaryon:
Adds support to resolve the resources from nested stack (AWS::Cloudformation::Stack).


pkg/iac-providers/cft/v1/load-file.go
```diff
@@ -40,44 +42,96 @@ func (a *CFTV1) LoadIacFile(absFilePath string) (allResourcesConfig output.AllRe
...
+		return allResourcesConfig, err
+	}
+
+	// fill AllResourceConfigs
+	allResourcesConfig = make(map[string][]output.ResourceConfig)
+	var config *output.ResourceConfig
+	for _, resource := range configs {
+		config = &resource
+		config.Line = 1
+		config.Source = a.getSourceRelativePath(absFilePath)
```
patilpankaj212, line 57: should the `source` for the nested stack resources be the `template url` or the filepath?
@cesar-rodriguez @kanchwala-yusuf WDYT?

pkg/mapper/iac-providers/cft/config/cloudformation-stack.go
```diff
@@ -17,29 +17,47 @@
...
 	if len(s.NotificationARNs) > 0 {
 		cf.NotificationARNs = s.NotificationARNs
 	}
+
+	// Add and resolve template URL
 	if len(s.TemplateURL) > 0 {
 		cf.TemplateURL = s.TemplateURL
+
+		templateData, err := fn.DownloadBucketObj(s.TemplateURL)
+		if err == nil {
```
patilpankaj212, line 51: any error that occurred during the download of the s3 object is lost and would not be captured in the `directory scan errors`. We should capture these errors in `directory scan errors` along with logging.

pkg/mapper/iac-providers/cft/functions/s3-download.go
```diff
@@ -0,0 +1,161 @@
...
+		buf, err := downloadPublicTemplate(templateURL)
+		if err != nil {
+			zap.S().Debug("not a public object", zap.String("url", templateURL), err)
+		} else {
+			return buf, nil
+		}
+	}
+
+	// if not public get bucket name and key
+	s3c, err := NewS3Client()
```
patilpankaj212, line 86: Can this client not used for the public s3 object download? 
Just wanted to understand why we require a separate download path for public templates.

sigmabaryon: Yeah, aws sdk perfiorms bucket operations and if the AWS user doesnt have READ permissions for the bucket, they wont be able to Get the object, even if the object is public. Download for public templates is not a bucket operation.

pkg/mapper/iac-providers/cft/config/cloudformation-stack.go
sigmabaryon: Done this way to keep the violations in AWS::Cloudformation:: Stack intact. Propagating error will skip the violations for AWS::Cloudformation::Stack resource eveny time the nested template is not resolved.

pkg/iac-providers/cft/v1/load-file.go
kanchwala-yusuf: I believe, it should be the `template url//filepath`. This way the user can clearly identify the module and the file in which the resource has been initialized

cesar-rodriguez: I agree with @kanchwala-yusuf

patilpankaj212 (approved):

PR #953, opened 2021-07-30
Adds Terrascan pre-commit

mihirhasan:


.pre-commit-config.yaml
```diff
@@ -0,0 +1,5 @@
+repos:
```
cesar-rodriguez, line 1: We need this on the docs, but can't add it to the repo directly because it would conflict with our  tests.

cesar-rodriguez (approved):

Issue #957, opened 2021-08-04
AC_K8S_0131 triggers on a Namespace resource


danmx:
* terrascan version:
* Operating System: 

### Description

I'm using latest Terrascan Github Action and it triggers `AC_K8S_0131` on `kubernetes_namespace` even though I have `namespace: whatever` set in in my `kustomization.yaml` and a Namespace resource is not a namespaced one.

```
 	Description    :	Ensure default name space is not in use in Kubernetes Namespace
	File           :	kustomization.yaml
	Line           :	1
	Severity       :	LOW
	Rule Name      :	defaultNamespaceUsed
	Rule ID        :	AC_K8S_0131
	Resource Name  :	valheim
	Resource Type  :	kubernetes_namespace
	Category       :	Security Best Practices
```

The issue seems to be here: https://github.com/accurics/terrascan/blob/2fdc14d96fa82cb114ab30c4d1a0c6294f6b77ad/pkg/policies/opa/rego/k8s/kubernetes_namespace/defaultNamespaceUsed.rego#L8

### What I Did

https://github.com/danmx/valheim-server-k8s/runs/3245324018?check_suite_focus=true

PR #961, opened 2021-08-04
fix(policies): removing false-positive for K8s namespaces

danmx:
Removing AC_K8S_0131 which is triggered for Kubernetes Namespaces.
It should address #957.

Signed-off-by: danmx <daniel@iziourov.info>


harkirat22:
Thanks @danmx for the fix. It looks good.

harkirat22 (approved):

PR #988, opened 2021-08-13
added binary based support for kustomize v2 and v3

nasir-rabbani:
* added **kutomize binary** based support for kustomize `v2` and `v3`.


pkg/iac-providers/kustomize/commons/load-dir.go
```diff
@@ -0,0 +1,163 @@
...
+
+	if err != nil {
+		err = fmt.Errorf("unable to read the kustomization file in the directory %s, error: %v", t.absRootDir, err)
+		zap.S().Error("error while reading the file", kustomizeFileName, zap.Error(err))
+		return allResourcesConfig, multierror.Append(t.errIacLoadDirs, results.DirScanErr{IacType: "kustomize", Directory: t.absRootDir, ErrMessage: err.Error()})
+	}
+
+	// ResourceConfig representing the kustomization.y(a)ml file
+	config := output.ResourceConfig{
+		Name:   filepath.Dir(t.absRootDir),
```
williepaul, line 74: Shouldn't these use relative paths?

pkg/iac-providers/kustomize/commons/load-dir.go
```diff
@@ -0,0 +1,163 @@
...
+
+		cmd := exec.Command(exe, arg1, arg2)
+
+		var stdOut, stdErr bytes.Buffer
+		cmd.Stdout = &stdOut
+		cmd.Stderr = &stdErr
+
+		err := cmd.Run()
+
+		if stdErr.String() != "" {
```
williepaul, line 133: Checking for data in stdErr and exiting could cause failures when there are none... Log data is often written to stderr, for example. Recommend removing this and just relying on the error code (next if statement)

pkg/iac-providers/kustomize/v3/testdata/simple-deployment/deployment.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: apps/v1
```
williepaul, line 1: One suggestion--can we keep any tests common for v2/v3/v4+ in a common dir, but leave any version-specific ones in the v2/v3/v4 dirs? This would just help keep maintenance tasks down

pkg/iac-providers/kustomize/commons/load-dir.go
nasir-rabbani: Thanks for pointing this out. I am using stdErr to get the `detailed error cause`, as `error` was returning only exit code. But yeah we should not return just because we get the stdErr.

pkg/iac-providers/kustomize/commons/load-dir.go
nasir-rabbani: This implementation was there from before and I have kept it as it was.

pkg/iac-providers/kustomize/v4/load-file.go
```diff
@@ -1,18 +1,11 @@
...
-
-var (
-	errLoadIacFileNotSupported = fmt.Errorf("load iac file is not supported for kustomize")
 )
 
 // LoadIacFile is not supported for kustomize. Only loading directories that have kustomization.y(a)ml file are supported
 func (k *KustomizeV4) LoadIacFile(absRootPath string, options map[string]interface{}) (allResourcesConfig output.AllResourceConfigs, err error) {
-	zap.S().Error(errLoadIacFileNotSupported)
-	return make(map[string][]output.ResourceConfig), errLoadIacFileNotSupported
+	return commons.NewKustomizeDirectoryLoader(absRootPath, options, false, versionSuffix).LoadIacFile()
```
patilpankaj212, line 10: since file loading is not supported, should we return `nil, commons.ErrLoadIacFileNotSupported` instead of creating a new KustomizeDirectoryLoader?

pkg/iac-providers/kustomize/v3/load-file.go
```diff
@@ -0,0 +1,11 @@
+package kustomizev3
+
+import (
+	"github.com/accurics/terrascan/pkg/iac-providers/kustomize/commons"
+	"github.com/accurics/terrascan/pkg/iac-providers/output"
+)
+
+// LoadIacFile is not supported for kustomize. Only loading directories that have kustomization.y(a)ml file are supported
+func (k *KustomizeV3) LoadIacFile(absRootPath string, options map[string]interface{}) (allResourcesConfig output.AllResourceConfigs, err error) {
+	return commons.NewKustomizeDirectoryLoader(absRootPath, options, true, versionSuffix).LoadIacFile()
```
patilpankaj212, line 10: since file loading is not supported, should we return `nil, commons.ErrLoadIacFileNotSupported` instead of creating a new KustomizeDirectoryLoader?

pkg/iac-providers/kustomize/commons/types.go
```diff
@@ -0,0 +1,32 @@
...
+// KustomizeDirectoryLoader implements kustomize directory/file loading
+type KustomizeDirectoryLoader struct {
+	absRootDir         string
+	options            map[string]interface{}
+	errIacLoadDirs     *multierror.Error
+	useKustomizeBinary bool
+	version            string
+}
+
+// KustomizeFileNames returns the valid extensions for k8s (yaml, yml, json)
```
patilpankaj212, line 26: `json` is mentioned but not returned. Should it be part of file names?

pkg/iac-providers/kustomize/commons/types.go
```diff
@@ -0,0 +1,32 @@
...
+	// YAMLExtension yaml
+	YAMLExtension = "yaml"
+	// YAMLExtension2 yml
+	YAMLExtension2 = "yml"
+	// KustomizeFileName kustomization
+	KustomizeFileName = "kustomization"
+)
+
+// KustomizeDirectoryLoader implements kustomize directory/file loading
+type KustomizeDirectoryLoader struct {
```
patilpankaj212, line 18: if this struct implements both directory and file loading should the type be `KustomizeLoader`?

pkg/iac-providers/kustomize/commons/load-file_test.go
```diff
@@ -13,22 +13,22 @@ func TestLoadIacFile(t *testing.T) {
...
-			kustomize: KustomizeV4{},
+			kustomize: KustomizeDirectoryLoader{},
 			wantErr:   errLoadIacFileNotSupported,
 		},
 	}
 
 	for _, tt := range table {
 		t.Run(tt.name, func(t *testing.T) {
-			_, gotErr := tt.kustomize.LoadIacFile(tt.filePath, tt.options)
+			_, gotErr := tt.kustomize.LoadIacFile()
```
patilpankaj212, line 31: since there is only one test and we are expecting a sentinel error back, should we just use `gotErr == errLoadIacFileNotSupported` instead of using `reflect.DeepEqual` ?

pkg/iac-providers/kustomize/commons/load-file.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+var (
+	errLoadIacFileNotSupported = fmt.Errorf("load iac file is not supported for kustomize")
+)
+
+// LoadIacFile is not supported for kustomize. Only loading directories that have kustomization.y(a)ml file are supported
+func (t KustomizeDirectoryLoader) LoadIacFile() (allResourcesConfig output.AllResourceConfigs, err error) {
+	zap.S().Error(errLoadIacFileNotSupported)
+	return make(map[string][]output.ResourceConfig), errLoadIacFileNotSupported
```
patilpankaj212, line 17: should we return `nil, errLoadIacFileNotSupported`?

pkg/iac-providers/kustomize/commons/load-file.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"fmt"
+
+	"github.com/accurics/terrascan/pkg/iac-providers/output"
+	"go.uber.org/zap"
+)
+
+var (
+	errLoadIacFileNotSupported = fmt.Errorf("load iac file is not supported for kustomize")
```
patilpankaj212, line 11: I think this error can be exported and used in other packages as well !!

patilpankaj212 (changes requested):

patilpankaj212 (dismissed):

patilpankaj212 (dismissed):

patilpankaj212 (approved):

cesar-rodriguez (changes requested):
Would you mind rebasing with master?

cesar-rodriguez (approved):

PR #994, opened 2021-08-16
terrascan should exit with non zero exit code when scan error are present

patilpankaj212:
1. exit with 5 when scan errors are present with violations
2. exit with 4 when scan errors are present without violations

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,8 +73,16 @@ To scan your code for security issues you can run the following (defaults to sca
 ```sh
 $ terrascan scan
 ```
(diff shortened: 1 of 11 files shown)


cesar-rodriguez (changes requested):
Would you mind updating the [Readme](https://github.com/accurics/terrascan/blob/master/README.md) and [docs](https://github.com/accurics/runterrascan.io/blob/main/content/en/docs/usage/command_line_mode.md#configuring-the-output-format-for-a-scan) explaining the new error codes?

cesar-rodriguez (approved):

PR #995, opened 2021-08-16
added support for argo-cd container image 

mihirhasan:


Rchanger:
I was not able to get it running. It was giving me error as `sh command not found for terrascan`

mihirhasan:
> I was not able to get it running. It was giving me error as `sh command not found for terrascan`

Can you specify what command caused you to run into this error? I suspect it has to do with execute permissions.

Rchanger:
> > I was not able to get it running. It was giving me error as `sh command not found for terrascan`
> 
> Can you specify what command caused you to run into this error? I suspect it has to do with execute permissions.

```
Error: unknown command "sh" for "terrascan"
Run 'terrascan --help' for usage.
```
Got this error in the pre-sync hook.

build/argo-Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+
+RUN chmod u+x /home/terrascan/bin/terrascan-remote-scan.sh
+
+# run as non root user
+USER 101
+
+ENV PATH /go/bin:$PATH
+
+# copy terrascan binary from build
+COPY --from=builder /go/bin/terrascan /go/bin/terrascan
```
Rchanger, line 48: I don’t think we need to build the terrascan binary again we just need a simple docker image that will have our script to connect with the already running terrascan pod. So you can create a simple docker image that will have our **terrascan-remote-scan.sh** script and the ability to run a curl command.

PR #992, opened 2021-08-16
update resource type to map[string]bool

patilpankaj212:
1. changed `resource` type in metadata json from `string` to `map[string]bool` and made related changes in resource type filter specification.
2. marked a few e2e tests pending.

**pending** - add more items in resource types of a metadata json if applicable.

PR #1001, opened 2021-08-20
Revert "update resource type to map[string]bool"

cesar-rodriguez:
Reverts accurics/terrascan#992 for now as it would break people using older versions of Terrascan


patilpankaj212 (approved):
e2e tests are failing due to latest policies in the master branch, can be fixed in next PR.

PR #1003, opened 2021-08-24
extract stage name of image in Dockerfile

patilpankaj212:
docker FROM can have an optional image name, which can be specified using AS
eg: FROM ubuntu:latest AS builder


patilpankaj212:
this can be added later if required.

PR #1037, opened 2021-10-01
Optimized disk encryption check code.

IrinaRaza:
Created small functions for encryption settings check.


gaurav-gogia (approved):
Looks good ✌️

PR #1105, opened 2021-12-31
fix input CFT templates based on goformation structs

patilpankaj212:


kanchwala-yusuf (approved):
LGTM,
nit: we need better comments for the `fixWithType` and `examineStruct` methods. Also, methods can be associated with the `CFTV1{}` struct

Issue #1122, opened 2022-01-21
Issue with skipped violations using CircleCI and JUnit output format


gchappel:
* terrascan version: v1.8.0
* Operating System: Linux

### Description

I/we are running Terrascan tests against some Terraform code, but trying to skip one known validation. This is executing through CircleCI, with the output in JUnit XML format and stored for display in the CircleCI webUI using `store_test_results`

The issue is that skipped violations, when output in JUnit format contain both the `skipped` element AND still contain a `failure` element. This appears to be causing CircleCI to show skipped violations as test failures, meaning that:

* on a successful CircleCI run (where `terrascan` exits with `0`) the Tests tab on the CircleCI build still shows errors where there should be none.
* on a failed CircleCI run (where `terrascan` exits with `>0`) the Tests tab shows all skipped violations as errors, making it difficult to pinpoint what _actually_ needs fixing at first glance. For example if you have 9 known violations with skip tags, and 1 genuine violation, they are all displayed identically as failed tests in the CircleCI UI.

I can't share the Terraform code itself but if required I'm happy to put together a minimal reproduction for this

I believe the issue is:
* The [JUnit spec](https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd) allows only **one of** the skipped/failure/error etc elements to exist (because these are all contained within an `<xs:choice>` element
* https://github.com/accurics/terrascan/blob/master/pkg/writer/junit_xml.go#L166-L169 says that if something is marked as skipped, it should have both a `skipped` AND a `failure` element, which appears to be against this spec
* CircleCI's JUnit parser is displaying all `testcase`s that have a `failure` element, without taking the `skipped` element into account as the use of these two elements together appears to be outside the schema

### What I Did

```
Paste the command(s) you ran and the output.
If there was a crash, please include the traceback here.
```

PR #1123, opened 2022-01-21
Prevent adding both skipped and failure elements

gchappel:
This is a potential fix to prevent both `skipped` and `failure` elements on a skipped violation in JUnit output format

Fixes: #1122

### DIFF
--- a/pkg/writer/junit_xml.go
+++ b/pkg/writer/junit_xml.go
@@ -164,18 +164,24 @@ func violationsToTestCases(violations []*results.Violation, isSkipped bool) []JU
 	for _, v := range violations {
 		var testCase JUnitTestCase
 		if isSkipped {
-			testCase = JUnitTestCase{Failure: new(JUnitFailure), SkipMessage: new(JUnitSkipMessage)}
-			testCase.SkipMessage.Message = v.Comment
+			testCase = JUnitTestCase{SkipMessage: new(JUnitSkipMessage)}
+			// since junitXML doesn't contain the attributes we want to show as violations
+			// we would add details of violations in the skip message, with any provided skip comment
+			if v.Comment == "" {
+				testCase.SkipMessage.Message = getViolationString(*v)
+			} else {
+				testCase.SkipMessage.Message = fmt.Sprintf("(skip comment: %s) %s", v.Comment, getViolationString(*v))
+			}
 		} else {
 			testCase = JUnitTestCase{Failure: new(JUnitFailure)}
+			// since junitXML doesn't contain the attributes we want to show as violations
+			// we would add details of violations in the failure message
+			testCase.Failure.Message = getViolationString(*v)
 		}
 		testCase.Classname = v.File
 		testCase.Name = fmt.Sprintf(testNameFormatFailed, v.ResourceName, v.LineNumber, v.RuleID)
 		testCase.Severity = v.Severity
 		testCase.Category = v.Category
-		// since junitXML doesn't contain the attributes we want to show as violations
-		// we would add details of violations in the failure message
-		testCase.Failure.Message = getViolationString(*v)
 		testCases = append(testCases, testCase)
 	}
 	return testCases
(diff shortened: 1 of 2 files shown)


patilpankaj212:
Hello @gchappel, thanks for creating the PR. Can you please fix the failing test?

pkg/writer/junit_xml.go
```diff
@@ -164,18 +164,24 @@ func violationsToTestCases(violations []*results.Violation, isSkipped bool) []JU
...
 		if isSkipped {
-			testCase = JUnitTestCase{Failure: new(JUnitFailure), SkipMessage: new(JUnitSkipMessage)}
-			testCase.SkipMessage.Message = v.Comment
+			testCase = JUnitTestCase{SkipMessage: new(JUnitSkipMessage)}
+			// since junitXML doesn't contain the attributes we want to show as violations
+			// we would add details of violations in the skip message, with any provided skip comment
+			if v.Comment == "" {
+				testCase.SkipMessage.Message = getViolationString(*v)
+			} else {
+				testCase.SkipMessage.Message = fmt.Sprintf("(skip comment: %s) %s", v.Comment, getViolationString(*v))
```
patilpankaj212, line 173: I think we should not add violation details in the skip message.
Does it help if the skip message contains violation information?

gchappel:
> Hello @gchappel, thanks for creating the PR. Can you please fix the failing test?

I'll take a look - it wasn't failing when I opened the PR but leave it with me.

pkg/writer/junit_xml.go
gchappel: I thought it would be useful to keep the failure reason; it would have been present before in the `<failure>` element, and I thought it would be the wrong move to throw it away (so the test can remain skipped but you can see _why_ it _would have_ failed in order to fix the issue before re-enabling the test.

If you'd rather not keep that context then I'm fine with that and can remove it.

gchappel:
I took a very quick look at the failing test, and it seemed weird that a test would fail when I only changed the behaviour of one specific output writer (and it's associated test).

So I checked out `master`, and the same tests are failing there too. I'm going to see if I can find a solution for it, but it may be the case that you need to get someone more familiar with the project to investigate this on `master` and then I can rebase this PR branch on that.

gchappel:
OK, I think I got it.

When I opened this PR (Jan 21st) I am 99% sure that `make unit-tests` was passing on my laptop. Between that time and the time the tests were run here, it appears that something changed which needed [this commit](https://github.com/accurics/terrascan/commit/52e4cdb27e246e2210417bd53b2f58a5279dd135) to address. When I re-tested `master` I was accidentally running from `master` in my own fork which was only up to date as of Jan 11th. Pulling `master` from origin allowed the unit tests to pass again, and rebasing this PR branch has fixed the tests here as well.

It seems like the tests may pull in external resources which are not under version control? It seems like there is some test data which changed and was no longer compatible with historical commit `b0259e8b0650a2a4b9a6a6f3b8ca31745cfa9515`, but only with `52e4cdb27e246e2210417bd53b2f58a5279dd135` and later.

gchappel:
Thanks @patilpankaj212 for re-running the tests; once again, the KaiMonkey project has moved on, and they fail. So I will rebase on `master` again, and re-iterate my request that these external artifacts are somehow versioned to avoid these spurious test failures in future.

pkg/writer/junit_xml.go
patilpankaj212: @gchappel I thought that this information would be misleading in the skip message, but if the information is useful, we can keep it.
We can remove it later if someone from the community requests for it.

patilpankaj212 (approved):

patilpankaj212:
Hello @gchappel, to track the issue with the artifacts being pulled from KaiMonkey, I have created a separate issue - https://github.com/accurics/terrascan/issues/1142. 

Thanks for your contribution.

pkg/writer/junit_xml.go
gchappel: > @gchappel I thought that this information would be misleading in the skip message, but if the information is useful, we can keep it. We can remove it later if someone from the community requests for it.

Oh, I already made the change to remove it based on your earlier comment, so the Git commit that you've approved and merged does _not_ have violation information in the skip message.

However, if the community asks for it to be _returned_ in future, then it's a very simple change which anyone can do based on the original commit from this PR.

gchappel:
> Hello @gchappel, to track the issue with the artifacts being pulled from KaiMonkey, I have created a separate issue - #1142.
> 
> Thanks for your contribution.

Perfect, hopefully this will make it easier for myself and/or others to contribute in future without having our PRs invalidated by a change outside of this repository in the test data; not to mention making the review process easier for yourselves.

PR #1136, opened 2022-02-02
add rule id to docs

jrobison-sb:
- I noticed a deprecation warning of `Deprecation warning : Use 'id' (AC_AWS_0058) instead of 'reference_id' (AWS.RDS.DataSecurity.High.0414)` and wanted to fix it in my config file, but then noticed that those ID's (`AC_AWS_0058`) aren't documented anywhere other than within the json of each rule. So I figured I would propose adding those ID's to the docs.
- I now see that maybe this repo isn't even the source of truth for the docs anymore? But the https://github.com/accurics/runterrascan.io repo doesn't seem to have any automation to rebuild these, or does it? Anyway, I'm happy to add this if possible but would appreciate guidance if I'm doing it in the wrong place.
- And this also fixes a bug in `docs/policies/document_policies.py` where it crashed when trying to iterate over files in `pkg/policies/opa/rego` when it was expecting to iterate over directories.

### DIFF
--- a/docs/policies/aws.md
+++ b/docs/policies/aws.md
@@ -1,420 +1,670 @@
 
 ### aws_iam_role_policy
-| Category | Resource | Severity | Description | Reference ID |
-| -------- | -------- | -------- | ----------- | ------------ |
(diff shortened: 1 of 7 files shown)


cesar-rodriguez (approved):

cesar-rodriguez:
Thanks @jrobison-sb ! 

@sangam14 we need an easier/better way to sync these with runterrascan.io

PR #1147, opened 2022-02-09
Add Policy AC_GCP_0296 - Service Attachment with ACCEPT_AUTOMATIC

hoexter:
GCP service attachments provide a way to share a single service
cross project and VPC boundaries. Since all examples show for ease
of use the connection preference `ACCEPT_AUTOMATIC` it's easy to
go to production with working but insecure setups.
Use `ACCEPT_MANUAL` and define the projects you would like to
allow to connect with a private service connect is the safe way.


cesar-rodriguez (approved):
Thanks!

PR #1148, opened 2022-02-10
update cft mapper

gaurav-gogia:
add support for:

1. app autoscaling policy
2. secrets manager rotation
3. iam group
4. ram resource share


patilpankaj212 (approved):
LGTM !!

PR #1199, opened 2022-03-24
update lambda function mapper

gaurav-gogia:
- improve lambda function mapper
- add support for `AWS::Serverless::Function`
- remove mapping for filename field
- remove source code hash field

### DIFF
--- a/pkg/mapper/iac-providers/cft/config/lambda-function.go
+++ b/pkg/mapper/iac-providers/cft/config/lambda-function.go
@@ -17,9 +17,6 @@
 package config
 
 import (
-	"strconv"
-	"strings"
-
 	"github.com/awslabs/goformation/v5/cloudformation/lambda"
 	"github.com/awslabs/goformation/v5/cloudformation/serverless"
 )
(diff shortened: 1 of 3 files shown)


pkg/mapper/iac-providers/cft/config/lambda-function.go
```diff
@@ -55,49 +54,102 @@ type LambdaFunctionConfig struct {
...
 // GetLambdaFunctionConfig returns config for LambdaFunction
-func GetLambdaFunctionConfig(f *lambda.Function) []AWSResourceConfig {
+func GetLambdaFunctionConfig(sf interface{}) []AWSResourceConfig {
+	if l, ok := sf.(*lambda.Function); ok {
+		return getLambdaConfig(l)
+	}
+
+	if s, ok := sf.(*serverless.Function); ok {
+		return getServerlessConfig(s)
+	}
```
patilpankaj212, line 64: if there are only two types that are going to call this function, we can re-write the function body as below:

```
if l, ok := sf.(*lambda.Function); ok {
  return getLambdaConfig(l)
}
return getServerlessConfig(s)
```

patilpankaj212 (approved):
LGTM !!

PR #1206, opened 2022-03-29
Update README.md

sangam14:
updated discord link !


Rchanger:
@sangam14 @cesar-rodriguez  It will look like below, we really want it to be like this
Screenshot 2022-03-30 at 7 25 07 PM

sangam14:
> @sangam14 @cesar-rodriguez It will look like below, we really want it to be like this
> Screenshot 2022-03-30 at 7 25 07 PM

its just easier to join discord via icon that work well I guess !

cesar-rodriguez (approved):

PR #1283, opened 2022-06-03
fixes: panic in case of zap logger init on windows os

Rchanger:


pkg/logging/logger.go
```diff
@@ -86,10 +94,16 @@ func GetLogger(logLevel, encoding, logDir string, encodingLevel func(zapcore.Lev
...
+		// so if we are on windows os register an sink which will handle the opening of file.
+		if utils.IsWindowsPlatform() {
+			zap.RegisterSink(terrascanZapSink, newTerrascanWinFileSink)
+			logDirPath = terrascanZapSinkWithSeparator + logDirPath
+		}
+		zapConfig.OutputPaths = append(zapConfig.OutputPaths, logDirPath)
 	}
-
-	// create zap logger
+	// should we panic here?
```
patilpankaj212, line 106: I think we should panic if we cannot create logger, or, propagate error back and exit with error message.

nasir-rabbani (approved):
LGTM

PR #1293, opened 2022-06-13
Adding CFT support for new resources

shreyas-phansalkar-189:
Adding CFT support for following resources:
- AWS Subnet
- AWS Nat Gateway
- AWS Route Table
- AWS Route Table Association
- AWS Route


patilpankaj212 (approved):
LGTM

PR #1319, opened 2022-07-08
container funcs to return dev friendly warnings

gaurav-gogia:


pkg/iac-providers/terraform/commons/extract-container-images.go
```diff
@@ -100,15 +100,15 @@ func fetchContainersFromAwsResource(resource jsonObj, hclBody *hclsyntax.Body, r
 			}
 			fileData, err := ioutil.ReadFile(fileLocation)
 			if err != nil {
-				zap.S().Errorf("error fetching containers from aws resource: %v", err)
+				zap.S().Warnf("error fetching containers from aws resource: %v", err)
```
nasir-rabbani, line 103: if it is just a warning, we probably shouldn't return the control. Also the messages clearly says `error`

pkg/iac-providers/terraform/commons/extract-container-images.go
```diff
@@ -126,7 +126,7 @@ func getContainersFromhclBody(hclBody *hclsyntax.Body) (results []output.Contain
 				for _, arg := range funcExp.Args {
 					re, diags := arg.Value(nil)
 					if diags.HasErrors() {
-						zap.S().Errorf("error fetching containers from aws resource: %v", getErrorMessagesFromDiagnostics(diags))
+						zap.S().Warnf("error fetching containers from aws resource: %v", getErrorMessagesFromDiagnostics(diags))
```
Rchanger, line 129: @gaurav-gogia let's change this msg as well, instead of `error fetching container`  make it `failed to fetch the container` since we are making it a warning keeping the error message will create confusion. WDYT

nasir-rabbani (approved):

PR #1373, opened 2022-08-25
fixed typo of the function name

Antsaniaina:
Fixed a typo in the Encryption Function name.


gaurav-gogia (changes requested):
Looks like some edge cases have been missed here
- encryption_settings is an array block
- when empty, encryption_settings may return config in this format `[{}]`

gaurav-gogia (approved):

PR #1398, opened 2022-09-20
Include DirScanErrors info in SARIF file

shaopeng-gh:
Hello, currently the DirScanErrors messages are not recorded in the SARIF file, 
this information could be useful for the SARIF file consumer.
This pr added this missing info so that the consumer of the SARIF file can have this info and act as necessary.
These messages are considered warning so the change is adding a new node to the json which should not break existing user.
We have updated the go-sarif sdk to the v2, which is necessary for this change.
And I see v2 have much more SARIF properties support than v1 which is great changes.
The json properties order changed slightly so the text cases are fixed accordingly.
Fixed a issue that multierr.Append() does not assign the result so it is missing from the object for later usage.
`golang.org/x/exp/typeparams` is removed by tidy mod.


gaurav-gogia (dismissed):
LGTM

shaopeng-gh:
@gaurav-gogia  thanks for the review.

shaopeng-gh:
@nasir-rabbani Let me know if looks good to merge, thanks.

shaopeng-gh:
note: The `build / validate (pull_request) Failing after 16m` happens after I fetch latest from `master`, I see other PRs have the same error so looks like not a issue of the PR.

shaopeng-gh:
Update: Test passed.

gaurav-gogia (approved):
LGTM

PR #1587, opened 2023-07-21
fix vulnerabilities reported by docker scan

nitumore:
http2 vulnerability reported by docker scan fix


nitumore:
throwing crash so revering this

PR #1730, opened 2025-04-03
Update README.md

Adesoji1:
Removed the $ and included sudo command in the command line that causes error when you copy and paste into the terminal to install terrascan

Before:

image

After:
image


Adesoji1 (commented):
Screenshot from 2025-04-03 19-32-02

nmoretenable (approved):

